Issue #3, opened 2022-10-12
Setting `resource.name` via `datadog_tag` does not seem to have any effect


mcuelenaere:
I would expect the following configuration to work:

```nginx
location / {
  proxy_pass http://api_node;
  datadog_tag resource.name "$request_method $uri";
}
```

However, in practice, this does not override the [built-in `resource.name` variable](https://github.com/DataDog/nginx-datadog/blob/535a291ce96d8ca80cb12b22febac1e138e45847/src/tracing_library.cpp#L201).

I suspect this is because `add_datadog_tag()` does not overwrite tags, but rather [adds them to the end of the `tags` array](https://github.com/DataDog/nginx-datadog/blob/4c2790ce6bb3b6830c128f2c21ac9232671fa9ab/src/datadog_directive.cpp#L145).

Although OTOH, when reading through the code, it looks like the spans are [being set by iterating over this array](https://github.com/DataDog/nginx-datadog/blob/4c2790ce6bb3b6830c128f2c21ac9232671fa9ab/src/request_tracing.cpp#L41-L43). So maybe the `SetTag()` function is treating `resource.name` in a special way and/or setting tags multiple times is not allowed?

In any case, I would like to override the default `resource.name` tagging strategy and at the moment this seems to be impossible.

dgoffredo:
Thanks for describing the problem.  I'll look into this.

dgoffredo:
Right you are, I've reproduced the issue.  Searching for a fix.

dgoffredo:
`unordered_map` is... unordered.

https://github.com/DataDog/nginx-datadog/blob/535a291ce96d8ca80cb12b22febac1e138e45847/src/ngx_http_datadog_module.cpp#L513

Embarrassing, but at least the fix is simple.

dgoffredo:
I just got back from a three week vacation, and this issue is not on the top of my TODO list, but I am still aware of this problem and will explore solutions in due time.

dgoffredo:
With the [most recent release][1], you can set the resource name using the dedicated `datadog_resource_name` directive:
```nginx
location / {
  proxy_pass http://api_node;
  datadog_resource_name "$request_method $uri";
}
```
However, that won't be necessary, because the default resource name has been changed to `$request_method $uri`.

[1]: https://github.com/DataDog/nginx-datadog/releases/tag/v0.4.0

PR #8, opened 2022-12-08
Revise resource names:

dgoffredo:
- Add configuration directive `datadog_resource_name` for setting the resource name of the request span.
- Add configuration directive `datadog_location_resource_name` for setting the resource name of the location span.
- Change the default resource name to be `$request_method $uri` instead of `$request_method $datadog_location`.

This will address https://github.com/DataDog/nginx-datadog/issues/3 and a related Support ticket.


cgilmour (approved):
Looks good!

PR #16, opened 2023-03-02
make builds compatible with MacOS

aetimmes:
There are currently a couple issues with running `make build-in-docker` on MacOS:

* passing `$(MAKE)` into `run_in_build_image.sh` ends up trying to run `/Library/Developer/CommandLineTools/usr/bin/make` in a Linux image, which fails
* the `sed` commands in `bin/nginx_release_downloads.sh` don't work with BSD sed, so we call GNU sed
* MacOS `tty` has `-s` but not `--silent`

This PR addresses those issues while (I think) maintaining Linux compatibility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ clobber: clean
 
 .PHONY: build-in-docker
 build-in-docker: sources
-	bin/run_in_build_image.sh $(MAKE) BUILD_DIR=.docker-build build
+	bin/run_in_build_image.sh make BUILD_DIR=.docker-build build
 
 .PHONY: test
 test: build-in-docker
--- a/Makefile
+++ b/Makefile
@@ -70,8 +70,9 @@ clean:
 .PHONY: clobber
 clobber: clean
 	rm -rf \
-	    nginx \
-		dd-opentracing-cpp/deps
+		nginx \
+		dd-opentracing-cpp/deps \
+		dd-opentracing-cpp/CMakeFiles
 
 .PHONY: build-in-docker
 build-in-docker: sources
--- a/bin/nginx_release_downloads.sh
+++ b/bin/nginx_release_downloads.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 # Print lines:
 #
@@ -14,7 +14,7 @@
 
 downloads=https://nginx.org/download
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [ "$(uname)" = 'Darwin' ]; then
   SED='gsed'
 else
   SED='sed'
--- a/bin/nginx_release_downloads.sh
+++ b/bin/nginx_release_downloads.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Print lines:
 #
@@ -14,6 +14,12 @@
 
 downloads=https://nginx.org/download
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  SED='gsed'
+else
+  SED='sed'
+fi
+
 if [ $# -gt 1 ]; then
     >&2 printf 'At most one argument supported, but %d were specified.\n' $#
     exit 1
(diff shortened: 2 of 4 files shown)


dgoffredo:
Thanks for contributing this. I no longer have access to a Mac, but my teammate does.

There are a few things I have thoughts about, so I'll be leaving comments inline.

Makefile
```diff
@@ -70,12 +70,13 @@ clean:
...
-	    nginx \
-		dd-opentracing-cpp/deps
+		nginx \
+		dd-opentracing-cpp/deps \
+		dd-opentracing-cpp/CMakeFiles
 
 .PHONY: build-in-docker
 build-in-docker: sources
-	bin/run_in_build_image.sh $(MAKE) BUILD_DIR=.docker-build build
+	bin/run_in_build_image.sh make BUILD_DIR=.docker-build build
```
dgoffredo, line 79: `$(MAKE)` is the GNU Make way of doing recursive make, per <https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html>.

You explained:

> passing $(MAKE) into run_in_build_image.sh ends up trying to run /Library/Developer/CommandLineTools/usr/bin/make in a Linux image, which fails

I wonder if the issue is the difference between GNU Make and non-GNU make. Maybe if instead of this change, we specified `gmake` everywhere we currently have `make`, it would work. What do you think?

bin/nginx_release_downloads.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
dgoffredo, line 1: These scripts run in environments without `bash` (such as in Alpine), so I keep it POSIX.

bin/nginx_release_downloads.sh
```diff
@@ -14,22 +14,28 @@
 
 downloads=https://nginx.org/download
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  SED='gsed'
+else
+  SED='sed'
+fi
```
dgoffredo, line 21: Per my comment above, consider this instead:
```shell
if [ "$(uname)" = 'Darwin' ]; then
  SED='gsed'
else
  SED='sed'
fi
```

bin/nginx_release_downloads.sh
```diff
@@ -14,22 +14,28 @@
...
+
 if [ $# -gt 1 ]; then
     >&2 printf 'At most one argument supported, but %d were specified.\n' $#
     exit 1
 fi
 
 filter() {
     if [ $# -eq 1 ]; then
-        grep "^$(echo "$1" | sed 's/\./\\./g')[. ]" | tail -1 | awk '{print $2}'
+        grep "^$(echo "$1" | $SED 's/\./\\./g')[. ]" | tail -1 | awk '{print $2}'
```
dgoffredo, line 30: `shellcheck` will want `"$SED"` here and below.

dgoffredo (changes requested):
Thanks again for patching this up.

See my comments about GNU make, and about sh vs. bash.

I'll ask my teammate to try this out on his macbook.

Makefile
aetimmes: The two `make` binaries on my M1 MBP are both GNU Make:

```shell
$ /usr/bin/make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0

$ /Library/Developer/CommandLineTools/usr/bin/make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

Since the existing system already has GNU make, there's no gmake binary available on a default MacOS system or via homebrew.

Both `/usr/bin/make` and `/Library/Developer/CommandLineTools/usr/bin/make` expand `$(MAKE)` to `/Library/Developer/CommandLineTools/usr/bin/make`:

```shell
$ cat Makefile
default:
        echo $(MAKE)
$ /usr/bin/make
echo /Library/Developer/CommandLineTools/usr/bin/make
/Library/Developer/CommandLineTools/usr/bin/make
```

These two executables have different md5sums and inode numbers, so they're not the same file. The reason for the above behavior isn't clear to me, but I'm on a relatively clean MacOS install and this behavior is consistent across bash and zsh, so I believe this is the default behavior.

I'd also argue that what this particular script is doing isn't (or perhaps shouldn't be) truly recursive make, as it is executing `make` in what is essentially an entirely new userspace inside the build container.

bin/nginx_release_downloads.sh
aetimmes: Ack, addressed in https://github.com/DataDog/nginx-datadog/pull/16/commits/57c3eedde296169327ad1d20b9b2d395b073a086

bin/nginx_release_downloads.sh
aetimmes: Ack, addressed in https://github.com/DataDog/nginx-datadog/pull/16/commits/57c3eedde296169327ad1d20b9b2d395b073a086

bin/nginx_release_downloads.sh
aetimmes: Shellcheck (0.9.0 on my machine) doesn't complain about this since we explicitly set `$SED` to one of two strings without whitespace earlier in the script.

aetimmes:
Thanks for taking a look at this. I was AFK for the last week or so, but took the time to address the comments above; LMK your thoughts.

dgoffredo:
Moved to https://github.com/DataDog/nginx-datadog/pull/28.

PR #17, opened 2023-03-14
eliminate segfaults in the case of nonexistent http configs

aetimmes:
Currently, if nginx is started with a config that only loads ngx_http_datadog_module.so and doesn't include a http config stanza, nginx will segfault in the module initializer. This PR ensures that in the above case, nginx will exit after producing a loggable error with human-parsable output.

### DIFF
--- a/src/ngx_http_datadog_module.cpp
+++ b/src/ngx_http_datadog_module.cpp
@@ -207,7 +207,7 @@ static ngx_command_t datadog_commands[] = {
       NGX_HTTP_LOC_CONF_OFFSET,
       0,
       nullptr),
-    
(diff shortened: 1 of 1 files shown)


dgoffredo:
Thanks for the patch. I'm testing it now.

This PR didn't trigger anything in CircleCI. I wonder why.

dgoffredo:
I've proposed <https://github.com/DataDog/nginx-datadog/pull/18> to address this issue instead.

Thanks again for finding this and offering a fix.

aetimmes:
Works for me. Thanks for taking a look!

PR #18, opened 2023-03-14
no http block in conf → no tracing

dgoffredo:
This addresses the crash for which a fix is proposed in <https://github.com/DataDog/nginx-datadog/pull/17>, but in a different way.

Currently, if the `http` block is missing from nginx's configuration, our module crashes on load.

Andrew's patch instead returns an error on load.

This patch instead puts the module into a no-op mode. I like this solution better, because the absence of an `http` block means that there is no behavior for us to alter (nginx-datadog is an _http_ module).

### DIFF
--- a/bin/release.py
+++ b/bin/release.py
@@ -283,7 +283,8 @@ def prepare_release_artifact(build_job_number, work_dir):
             f"Job number {build_job_number} doesn't have an 'ngx_http_datadog_module.so' build artifact."
         )
 
-    nginx_version_info = urllib.request.urlopen(nginx_version_info_url).read().decode('utf8')
+    nginx_version_info = urllib.request.urlopen(
+        nginx_version_info_url).read().decode('utf8')
     module_path = work_dir / 'ngx_http_datadog_module.so'
     download_file(module_url, module_path)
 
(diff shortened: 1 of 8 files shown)


bin/release.py
```diff
@@ -283,7 +283,8 @@ def prepare_release_artifact(build_job_number, work_dir):
             f"Job number {build_job_number} doesn't have an 'ngx_http_datadog_module.so' build artifact."
         )
 
-    nginx_version_info = urllib.request.urlopen(nginx_version_info_url).read().decode('utf8')
+    nginx_version_info = urllib.request.urlopen(
+        nginx_version_info_url).read().decode('utf8')
```
dgoffredo, line 287: This is from the code formatter.

test/cases/orchestration.py
```diff
@@ -323,7 +323,8 @@ class Orchestration:
     """
 
     def __init__(self):
-        print('The test runner is running in the following environment:', os.environ)
+        print('The test runner is running in the following environment:',
+              os.environ)
```
dgoffredo, line 327: This is from the code formatter.

dgoffredo:
Caleb and I looked over this, and I'm going to look more closely at which of these added branches are necessary. Some of them are probably dead code.

After that, we should be good to merge.

PR #24, opened 2023-05-05
support new versions of nginx and amazonlinux

dgoffredo:


dgoffredo:
These won't be available until I cut a release. I'll cut a minor/patch release early next week.

Issue #26, opened 2023-06-09
Error: "load_module" directive in nginx.conf not binary compatible


takashi-osako-compass:
I am encountering an issue with the "load_module" directive in my nginx.conf file while using the source code from tag v1.0.0 of the DataDog repository. I am running the code inside a Docker image based on `python:3.8.12-slim-buster`.

Here are the steps I followed:

1. Cloned the DataDog/nginx-datadog repository.
2. Checked out the `v1.0.0` tag.
3. Executed the `bin/install_build_tooling.sh` script to install the necessary build tooling.
4. Created a `nginx-version-info` file and set `NGINX_VERSION=1.14.2` in it.
5. Ran `make build` to build the code successfully.
6. Updated the nginx.conf file to include the required module.

Attempted to start the application, but encountered the following error:
`nginx: [emerg] module "libngx_http_datadog_module.so" is not binary compatible in /docker/webapp/nginx.conf:2`

I have double-checked the syntax and path to the module file, ensuring they are correct. However, the error persists.  I also removed `--with-compat` option in Makefile to see if it changes, but the error still persists.

Could you please provide guidance on how to troubleshoot and resolve this error based on the procedures I followed?

dgoffredo:
> I also removed --with-compat option in Makefile to see if it changes, but the error still persists.

Good idea, something like that (an nginx `./configure` flag) probably at the heart of the issue.

Which nginx installation are you using within `python:3.8.12-slim-buster`? Is it the one available via `apt-get`? I'll pull that image now and look around...

If I `apt update` and `apt install nginx` within that python image, and then `nginx -V` with some added newlines, I get:
```
nginx version: nginx/1.14.2
built with OpenSSL 1.1.1n  15 Mar 2022 (running with OpenSSL 1.1.1d  10 Sep 2019)
TLS SNI support enabled
configure arguments:
 --with-cc-opt='-g -O2 -fdebug-prefix-map=/build/nginx-FlcIR2/nginx-1.14.2=. -fstack-protector-strong -Wformat -Werror=format-security -fPIC -Wdate-time -D_FORTIFY_SOURCE=2'
 --with-ld-opt='-Wl,-z,relro -Wl,-z,now -fPIC'
 --prefix=/usr/share/nginx
 --conf-path=/etc/nginx/nginx.conf
 --http-log-path=/var/log/nginx/access.log
 --error-log-path=/var/log/nginx/error.log
 --lock-path=/var/lock/nginx.lock
[...]

PR #29, opened 2023-06-16
Dynamically Detect Nginx Parameters and Install Additional Packages for Successful Compilation with Image

takashi-osako-compass:
This pull request aims to enhance the installation process of the module by dynamically reading the parameters used to build Nginx if it is already installed. By doing so, we ensure that the module is compiled with the same parameters, promoting compatibility and consistency.

To achieve this, the following changes have been made:

1. Introduce a parameter detection mechanism: The installation script now checks if Nginx is installed and retrieves its build parameters dynamically. This ensures that the module is compiled using the same options, flags, and configuration as the existing Nginx installation.

1. Add compatibility with the Buster image: In order to compile successfully with the image, additional packages have been included as dependencies. These packages are now automatically installed during the compilation process, guaranteeing a smooth and error-free build.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,6 +38,7 @@ jobs:
       MAKE_JOB_COUNT: 8
     steps:
     - checkout
+    - run: bin/install_build_tooling.sh
     - run: printf 'NGINX_VERSION=%s\n' << parameters.nginx-version >> >>nginx-version-info
     - run: printf 'BASE_IMAGE=%s\n' '<< parameters.base-image >>' >>nginx-version-info
     - run: make build
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ MAKE_JOB_COUNT ?= $(shell nproc)
 
 .PHONY: build
 build: build-deps nginx/objs/Makefile sources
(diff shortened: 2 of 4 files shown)


dgoffredo:
Thank you for proposing these changes, @takashi-osako-compass!

I'm glad to see that you [got it working][1].

My teammate and I had been discussing how we ought to support build modes where the `./configure` arguments differ from what we currently use. We imagined a "build telemetry script" that would detect and nginx installation and collect information about it, like you have done here. Additionally, we wanted to include information about the host system, standard c library, etc.

I won't merge these changes exactly as you've proposed them, because there is overlap with the solution that we will come up with. Also, there are a few small details that I would change. I'll leave this pull request open until this "build telemetry script" is written.

Hopefully by the next release, you will be able to build the module within your image without having to maintain a patch like this.

Thanks again for the contribution.

[1]: https://github.com/DataDog/nginx-datadog/issues/26#issuecomment-1593998581

takashi-osako-compass:
Thank you for your feedback! I'm thrilled to hear fix is on the way!

devopstravelwifi:
What's the solution for this?

PR #35, opened 2023-08-17
increase portability of tests

dgoffredo:
I was working on a side project last weekend that had me running `make test` on my personal laptop. It failed.

The reason was that `docker-compose` on that laptop prefixes messages like "Creating container [...] ..." with a space, so it's " Creating container [...] ...". `orchestration.py` was thus failing to recognize such lines and indicating that a container was created, and so the test driver was spinning forever waiting for things to come up.

This revision fixes that by adding the prefix `\s*` to all relevant regex patterns.

Another issue was how `docker-compose ps` works. The tests were written to expect that when `docker-compose ps <service>` has nothing to say, then the command will exit with a nonzero status. On my personal laptop, this is not true. So, now we wait for the condition "nonzero exit status and nonempty output." Five retries is no longer enough, so I increased it to 100.

Finally, I changed all mentions of `docker-compose` to be instead `docker compose`. They are different, though on CircleCI I think that `docker-compose` was just a wrapper that forwarded to `docker compose`. The original Python `docker-compose` is completely replaced by Docker's "compose" plugin (`docker compose`, written in Go).

The tests now pass on my personal laptop. Let's see what CircleCI thinks of these changes.

### DIFF
--- a/test/.gitignore
+++ b/test/.gitignore
@@ -1,3 +1,2 @@
 __pycache__/
-docker-compose-verbose.log
 test.times
--- a/test/README.md
+++ b/test/README.md
@@ -1,7 +1,7 @@
 Tests
 =====
 These are integration tests.  They're written in python, and use
-`docker-compose` to orchestrate an instance of nginx containing the module
+`docker compose` to orchestrate an instance of nginx containing the module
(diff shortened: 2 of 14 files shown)


dgoffredo:
thanks, Caleb

PR #40, opened 2023-09-05
Update macrobenchmarks

ddyurchenko:
1. Replace ben-runner with bp-runner.
2. Set k6 rate to more representative values.
3. Ensure job artifacts are uploaded even if job fails.
4. Run macrobenchmark on every master branch commit.


igoragoli (approved):

dgoffredo (approved):
Thanks for the tweaks!

lgtm_stamp_by_nushe4ka_d5w539d-fullview

PR #44, opened 2023-10-18
work around DNS delays in ARM64 CI

dgoffredo:
nginx uses `gethostbyname()` to look up IP addresses for upstreams (unless you designate a resolver). Both libc flavors (glibc and musl) as configured in our test images, will first check `/etc/hosts` before going to DNS. DNS sometimes takes five seconds (instead of the more typical < 1 ms), which causes a timeout in the tracing library, which causes tests to fail that are checking whether the tracer sent a trace to the Agent.

This revision pre-populates the nginx container's `/etc/hosts` with the other `docker compose` services' IP addresses.


test/cases/orchestration.py
```diff
@@ -325,6 +329,30 @@ def header_args():
...
+    environment in CircleCI, we started seeing intermittent delays in DNS
+    lookups, always about five seconds.
+
+    This function uses `getent` to look up the IP address of each `docker
+    compose` service, and then associates the address with the service name by
+    adding a line to /etc/hosts in the nginx service container. This way, the
+    default `gethostbyname()` resolver used by nginx will not use DNS."""
+    script = f"""
+    for service in {" ".join(f"'{service}'" for service in services)}; do
+        getent hosts "$service" >>/etc/hosts
```
dmehala, line 343: First time I encounter `getent`, interesting. I thought you would use `docker inspect`.

dgoffredo:
Thanks. I'll resolve the conflicts and merge after the ARM64 branch is merged into master.

dmehala:
It has been cherry-picked in #45

PR #47, opened 2023-11-14
[chore-ci] XML reporter

dmehala:
## Description

Quick improvement to store our test results in JUnit XML format instead of raw text. This greatly improves readability.

### Changes
- Add the [unittest-xml-reporting](https://github.com/xmlrunner/unittest-xml-reporting/tree/master/) module but with a small change to support `startTestRun` and `stopTestRun`.
- Update the CI to store the XML report and logs.
- Call `run.py` instead of `unittest` main module. ⚠️ There is a small regression here, for now `run.py` do not support filtering.

## Preview
Screenshot 2023-12-01 at 16 42 12


dgoffredo (changes requested):
Thanks for making these changes. The resulting "see failures in the CI" workflow is improved, and these changes address my "this sucks" comment in the CI config.

Per our discussion, the necessity to add the source of the XML reporter in tree makes these changes heavier than I would like. If we can `pip install pytest` and accomplish the same thing with much less source code, that would be preferable. Please investigate that before we decide whether to merge these changes.

dmehala:
Switch over Pytest.

PR #48, opened 2023-11-15
Generate 128-bit trace id

dmehala:
This draft PR is not intended to be merged.

PR #49, opened 2023-11-17
use dd-trace-cpp v0.1.12

dgoffredo:
This updates the version of `dd-trace-cpp` used, merges in Damien's work from #48, and formats the code.


dmehala (approved):
:shipit:

PR #51, opened 2023-11-21
chore: add architecture in release process

dmehala:
## Description

I iterated so much across multiple branches to add `arm64` support, and in the process, I inadvertently lost this particular change. The objective of this pull request is to introduce multi-architecture support through the release process.

## ⚠️ Note to reviewer
This change is not backward compatible. Should we care about backward compatibility?

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ include the following:
 
 Each release contains one zipped tarball per supported image above. The
 naming convention is
-`<base image with underscores><hyphen>ngx_http_datadog_module.so.tgz`,
-e.g. `nginx_1.23.1-alpine-ngx_http_datadog_module.so.tgz` or
-`amazonlinux_2.0.20230119.1-ngx_http_datadog_module.so.tgz`.
+`<base image with underscores>-<arch>-ngx_http_datadog_module.so.tgz`,
+e.g. `nginx_1.23.1-alpine-amd64-ngx_http_datadog_module.so.tgz` or
+`amazonlinux_2.0.20230119.1-arm64-ngx_http_datadog_module.so.tgz`.
 
 The zipped tarball contains a single file, `ngx_http_datadog_module.so`, which
 is the Datadog tracing nginx module.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -25,9 +25,9 @@ include the following:
 
 Each release contains one zipped tarball per supported image above. The
 naming convention is
-`<base image with underscores><hyphen>ngx_http_datadog_module.so.tgz`,
-e.g. `nginx_1.23.1-alpine-ngx_http_datadog_module.so.tgz` or
-`amazonlinux_2.0.20230119.1-ngx_http_datadog_module.so.tgz`.
+`<base image with underscores>-<arch>-ngx_http_datadog_module.so.tgz`,
```
dmehala, line 28: Subjective, I found `<hyphen>` was not as straightforward as `-` and could be misleading. What's your thought?

bin/release.py
```diff
@@ -296,8 +296,14 @@ def prepare_release_artifact(build_job_number, work_dir):
         raise Exception(
             f"BASE_IMAGE not found in nginx-version-info: {nginx_version_info}"
         )
+    if 'ARCH' not in variables:
```
dgoffredo, line 299: Please add a comment, either just above this `if` or as an additional paragraph after the comment above, saying what `ARCH` is, how it is used, and which values it can have.

README.md
dgoffredo: That's fine. `<hyphen>` I always had to read twice, like "what?"

My initial reasoning is that the "meta-syntax" `<thingy here>` uses angle brackets, and so using a hyphen literally might look like part of the meta-syntax instead of a literal hyphen.

Your way is better.

dgoffredo (changes requested):
Looks good, thanks for patching this.

Please add a section to the documentation (comments) at the top of `nginx-version-info.example` that exhaustively explains `ARCH`.

Other than that and another comment in the Python script (which you might choose still to omit), this looks good to me.

Once you've updated `nginx-version-info.example`, feel free to merge this.

dgoffredo (approved):
I'll merge this once the CI passes.

Backward incompatibility for release processes is okay, since we are the only ones cutting releases.

The fact that the naming convention for the `.tgz` files has changed could be problematic for [scripts][1] (or humans) that pull the latest release. We can break the scripts, but it's best not the break the humans. We've updated the relevant in-repo documentation, but still need to update the Datadog documentation (linked earlier). I think that's all we'll need to do; let me know if I missed something.

[1]: https://docs.datadoghq.com/tracing/trace_collection/proxy_setup/?tab=nginx#module-installation

PR #54, opened 2023-11-24
Add support for AL2023

gugaiz:
Add changes to work with AL2023 distributions

On this PR I am also changing the way it works with local images because the original file is creating a local tar but if we try to build with docker it fails because does not find the image on the local registry


dgoffredo:
Thanks for proposing this, @gugaiz.

@dmehala, please have a look at this so we can discuss it in our meeting tomorrow.

dgoffredo:
Hi again, @gugaiz. My teammate and I were looking at this today. We don't know when we'll be able to test support for this newer release of Amazon Linux, but we would like to support it. Based on your PR, it looks like it might not be too complicated.

The team is small and busy with other initiatives, but we intend to revise how our releases are managed, and so we will revisit this at that time.

dgoffredo:
Let's move this over to https://github.com/DataDog/nginx-datadog/pull/64. All that's left to do is build/push the ARM-based docker image.

PR #56, opened 2023-12-05
refactor: Implement simpler header injection mechanism

### LINKED ISSUE #19: "proxy_set_header" doesn't set header to upstream server.
This issue is similar to #5 but issue #5 was closed without resolution.

When enabling APM in Nginx a Header set in the `proxy_set_header` directive doesn't seem to be set in the request to the upstream server. We try to set a `Forwarded` header to the upstream server using `proxy_set_header Forwarded $proxy_add_forwarded;` but our upstream never receive the value in the header.

Configuration
OS: `alpine`
nginx: `v1.23.3`
nginx-datadog: `v0.5.3`
package: `nginx_1.23.3-alpine-ngx_http_datadog_module.so.tgz`

nginx.conf
```
user  nginx;

load_module /usr/lib/nginx/modules/ngx_http_datadog_module.so;

worker_processes  auto;

# error_log stderr notice; # To Do - Change back to Info for Prod
error_log stderr info;
pid        /var/run/nginx.pid;

worker_rlimit_nofile 16384;

events {
  worker_connections  8192;
}

http {
    server_tokens       off;                # ensures the version information is not returned
    keepalive_timeout   10s;                # CIS: set to '10' or less, but not 0. - Setting a keep-alive timeout on the server side helps mitigate denial of service attacks that establish too many persistent connections, exhausting server resources.
    send_timeout        10s;                # CIS: set to '10' or less. The send_timeout directive sets a timeout for transmitting a response to the client between two successive write operations. Setting the send_timeout directive on the server side helps mitigate slow HTTP denial of service attacks by ensuring write operations taking up large amounts of time are closed.
    large_client_header_buffers 2 1k;       # CIS: 5.2.3 Ensure the maximum buffer size for URIs is defined
    add_header          Referrer-Policy "no-referrer";  # CIS: 5.3.4 Ensure the Referrer Policy is enabled and configured properly
    proxy_ssl_ciphers   EECDH:EDH:!LOW:!NULL:!aNULL:!EXP:!ADH:!SSLv2:!SSLv3:!TLSv1.0:!MD5:!RC4:!3DES:!IDEA;

    gzip                on;                 # Enables gzipping of responses.
    gzip_proxied        any;                # enables compression for all proxied requests
    gzip_types          text/plain application/json;    # specify the content types which will be gzipped.

    sendfile            on;                 # This only concerns static files and only come into play when serving files from disk. http://nginx.org/en/docs/http/ngx_http_core_module.html#sendfile
    tcp_nopush          on;                 # This is only used when sendfile() is used and only concerns static files. http://nginx.org/en/docs/http/ngx_http_core_module.html#tcp_nopush
    add_header          Strict-Transport-Security "max-age=31536000; includeSubDomains";    # add HSTS header to all responsed to ensure the browser always connects over HTTPS - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

    real_ip_header      X-Forwarded-For;    # Defines the request header field whose value will be used as the client address.
    real_ip_recursive   on;                 # When more than one trusted IP is defined in 'set_real_ip_from', ignore them all and take the righ-most non-trusted IP.

    client_header_timeout 10s;              # If a client does not transmit the entire header within this time, the request is terminated with the 408 (Request Time-out) error.
    client_body_timeout   10s;              # Defines a timeout for reading client request body. The timeout is set only for a period between two successive read operations, not for the transmission of the whole request body. If a client does not transmit anything within this time, the request is terminated with the 408 (Request Time-out) error.
    client_max_body_size  8k;               # Sets the maximum allowed size of the client request body. If the size in a request exceeds the configured value, the 413 (Request Entity Too Large) error is returned to the client. Please be aware that browsers cannot correctly display this error. Setting size to 0 disables checking of client request body size.

    proxy_ssl_server_name on;               # Enables passing of the server name through TLS Server Name Indication extension (SNI, RFC 6066) when establishing a connection with the proxied HTTPS server.
    proxy_http_version    1.1;              # Sets the HTTP protocol version for proxying. By default, version 1.0 is used. Version 1.1 is recommended for use with keepalive connections and NTLM authentication.
    proxy_ssl_protocols   TLSv1.3 TLSv1.2;  # Only allowed protocols for upstream servers.
    proxy_connect_timeout 10s;              # Defines a timeout for establishing a connection with a proxied server.
    proxy_read_timeout    10s;              # Defines a timeout for reading a response from the proxied server. The timeout is set only between two successive read operations, not for the transmission of the whole response.
    proxy_send_timeout    10s;              # Sets a timeout for transmitting a request to the proxied server. The timeout is set only between two successive write operations, not for the transmission of the whole request. If the proxied server does not receive anything within this time, the connection is closed.
    proxy_set_header      Connection "";                              # removes the "close" value from the Connection header so connections to the upstream server can be reused.
    proxy_set_header      Host $proxy_host;                           # Set the `Host` header in the request to the upstream server.
    proxy_set_header      Forwarded $proxy_add_forwarded;             # Send the RFC7239 "Forwarded" header to upstream servers.
    proxy_set_header      X-Forwarded-For $proxy_add_x_forwarded_for; # Adds the client IP to the "X-Forwarded-For" header and send it to the upstream server.

    server_names_hash_bucket_size 128;   # Sets the bucket size for the server names hash tables.

    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    # ---- RFC7239 compliant Forwarded header - https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/ ----
    map $realip_remote_addr $proxy_forwarded_elem {
        # IPv4 addresses can be sent as-is
        ~^[0-9.]+$          "for=$realip_remote_addr;proto=$scheme;host=$host";

        # IPv6 addresses need to be bracketed and quoted
        ~^[0-9A-Fa-f:.]+$   "for=\"[$realip_remote_addr]\";proto=$scheme;host=$host";

        # Unix domain socket names cannot be represented in RFC 7239 syntax
        default             "for=unknown;proto=$scheme;host=$host";
    }

    map $http_forwarded $proxy_add_forwarded {
        # If the incoming Forwarded header is syntactically valid, append to it
        "~^(,[ \\t]*)*([!#$%&'*+.^_`|~0-9A-Za-z-]+=([!#$%&'*+.^_`|~0-9A-Za-z-]+|\"([\\t \\x21\\x23-\\x5B\\x5D-\\x7E\\x80-\\xFF]|\\\\[\\t \\x21-\\x7E\\x80-\\xFF])*\"))?(;([!#$%&'*+.^_`|~0-9A-Za-z-]+=([!#$%&'*+.^_`|~0-9A-Za-z-]+|\"([\\t \\x21\\x23-\\x5B\\x5D-\\x7E\\x80-\\xFF]|\\\\[\\t \\x21-\\x7E\\x80-\\xFF])*\"))?)*([ \\t]*,([ \\t]*([!#$%&'*+.^_`|~0-9A-Za-z-]+=([!#$%&'*+.^_`|~0-9A-Za-z-]+|\"([\\t \\x21\\x23-\\x5B\\x5D-\\x7E\\x80-\\xFF]|\\\\[\\t \\x21-\\x7E\\x80-\\xFF])*\"))?(;([!#$%&'*+.^_`|~0-9A-Za-z-]+=([!#$%&'*+.^_`|~0-9A-Za-z-]+|\"([\\t \\x21\\x23-\\x5B\\x5D-\\x7E\\x80-\\xFF]|\\\\[\\t \\x21-\\x7E\\x80-\\xFF])*\"))?)*)?)*$" "$http_forwarded, $proxy_forwarded_elem";

        # Otherwise, replace it
        default "$proxy_forwarded_elem";
    }
    # ---- END: RFC7239 compliant Forwarded header ----

    log_format json_custom escape=json
    '{'
    	...

        '"upstream.request.header.forwarded":"$proxy_add_forwarded",'       # "Forwarded" header sent to upstream server
        ...
    '}';

    access_log /dev/stdout json_custom;

    include /etc/nginx/conf.d/*.conf;
}

```

Proxy config
```
server {
    listen      80;
    server_name    example.com;

    large_client_header_buffers 4 8k;
    client_max_body_size          3m;

    proxy_connect_timeout 240s;
    proxy_read_timeout    240s;
    proxy_send_timeout    240s;
    keepalive_timeout     240s;
    send_timeout          240s;

    set $upstream_servers    example-internal.example.com;

    location / {
        proxy_pass https://$upstream_servers;

        proxy_redirect https://$proxy_host:443/ https://$host/;
    }
}
```

dmehala:
# Description
This commit introduces a refined header injection mechanism that modify the request structure with headers to inject, instead of mutating nginx's configuration with `proxy_set_header`, `fastcgi_param`, `grpc_set_header` or `uwsgi_param`.

Resolves #19

### Changes:
  - Add a DictWriter implementation for nginx.
  - Add a test covering issue #19.
  - Deprecate `opentracing_*_propagate_context`.
  - Remove tests no longer necessary.
  - Update existing tests to align with the new implementation.

## Note to reviewer
I am aware of the [OpenTracing saga with `headers_in`](http://mailman.nginx.org/pipermail/nginx-devel/2018-March/011008.html) and thruth be told, it made me a bit worry.

On the investigation front, OpenTracing's implementation wasn't exactly a walk in the park - memory lifecycle issues and not setting  the `lowercase_key` field [which could lead to a crash](https://www.nginx.com/resources/wiki/start/topics/examples/headers_management/). This implementation take care of both issues. 

This isn't emerging out of nowhere or against the trend; we have examples of plugins with comparable implementation:
- OpenTelemetry: https://github.com/open-telemetry/opentelemetry-cpp-contrib/blob/130c6e62ee6602b1fc034969a2156670910e37f1/instrumentation/nginx/src/propagate.cpp#L17
- Headers-more: https://github.com/openresty/headers-more-nginx-module/blob/3fb97bcf57b17106ee68f47eb01233cd0eeb2d54/src/ngx_http_headers_more_headers_in.c#L321

### DIFF
--- a/src/ngx_header_writer.h
+++ b/src/ngx_header_writer.h
@@ -1,0 +1,50 @@
+#pragma once
+
+#include "string_util.h"
+#include <datadog/dict_writer.h>
+
+extern "C" {
+#include <nginx.h>
+#include <ngx_config.h>
+#include <ngx_core.h>
+#include <ngx_http.h>
+}
+
+namespace datadog {
+namespace nginx {
+
+class NgxHeaderWriter : public datadog::tracing::DictWriter {
+  ngx_http_request_t *request_;
+  ngx_pool_t *pool_;
+
+public:
+  explicit NgxHeaderWriter(ngx_http_request_t *request)
+      : request_(request), pool_(request_->pool) {}
+
+  void set(std::string_view key, std::string_view value) override {
+    // Question: Should we check first if the key already exist?
+    const auto key_size = key.size();
+
+    ngx_table_elt_t *h = (ngx_table_elt_t*)ngx_list_push(&request_->headers_in.headers);
+    if (h == NULL) {
+      return;
+    }
+
+    h->hash = 1;
+
+    // HTTP proxy module expects the header to has a lowercased key value
+    // Instead of allocating twice the same key, `h->key` and `h->lowcase_key`
+    // use the same data.
+    h->key.data = (u_char*)ngx_pnalloc(pool_, sizeof(char) * key_size);
+    h->key.len = key_size;
+    for (std::size_t i = 0; i < key_size; ++i) {
+      h->key.data[i] = to_lower(key[i]);
+    }
+    h->lowcase_key = h->key.data;
+
+    h->value = to_ngx_str(pool_, value);
+  }
+};
+
+}
+}
--- a/src/datadog_directive.cpp
+++ b/src/datadog_directive.cpp
@@ -30,38 +30,6 @@ auto command_source_location(const ngx_command_t *command, const ngx_conf_t *con
                                           .directive_name = command->name};
 }
 
-// Dispatch to the "real" handler for the specified `command`, and then invoke
-// the specified `inject_propagation_commands` with the specified `cf`,
-// `command`, and `conf`.  `inject_propagation_commands` is intended to do the
-// Datadog-specific work associated with the hijacked `command`, e.g. insert
-// `proxy_set_header` directives into the current configuration context.
-// Return the value returned by `inject_propagation_commands`, or return
-// `NGX_CONF_ERROR` if an error occurs.
-char *hijack_pass_directive(char *(*inject_propagation_commands)(ngx_conf_t *cf,
-                                                                 ngx_command_t *cmd, void *conf),
-                            ngx_conf_t *cf, ngx_command_t *command, void *conf) noexcept try {
-  // First, call the handler of the actual command that we're hijacking, e.g.
-  // "proxy_pass".  Be sure to skip this module, so we don't call ourself.
-  const ngx_int_t rcode = datadog_conf_handler({.conf = cf, .skip_this_module = true});
-  if (rcode != NGX_OK) {
-    return static_cast<char *>(NGX_CONF_ERROR);
-  }
-
-  // Set the name of the proxy directive associated with this location.
-  if (const auto loc_conf = static_cast<datadog_loc_conf_t *>(
-          ngx_http_conf_get_module_loc_conf(cf, ngx_http_datadog_module))) {
-    loc_conf->proxy_directive = command->name;
-  }
-
-  // Second, call the Datadog-specific handler that sets up context
-  // propagation, e.g. `propagate_datadog_context`.
-  return inject_propagation_commands(cf, command, conf);
-} catch (const std::exception &e) {
-  ngx_log_error(NGX_LOG_ERR, cf->log, 0, "Datadog-wrapped configuration directive %V failed: %s",
-                command->name, e.what());
-  return static_cast<char *>(NGX_CONF_ERROR);
-}
-
 // An empty configuration instructs the member functions of `TracingLibrary` to
 // substitute a default configuration instead of interpreting the string as a
 // JSON encoded configuration.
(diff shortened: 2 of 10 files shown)


src/ngx_header_writer.h
```diff
@@ -0,0 +1,50 @@
...
+  void set(std::string_view key, std::string_view value) override {
+    // Question: Should we check first if the key already exist?
+    const auto key_size = key.size();
+
+    ngx_table_elt_t *h = (ngx_table_elt_t*)ngx_list_push(&request_->headers_in.headers);
+    if (h == NULL) {
+      return;
+    }
+
+    h->hash = 1;
```
dgoffredo, line 33: Please add a comment here explaining why (even if it's a link to some docs or a mailing list thread).

src/ngx_header_writer.h
```diff
@@ -0,0 +1,50 @@
...
+class NgxHeaderWriter : public datadog::tracing::DictWriter {
+  ngx_http_request_t *request_;
+  ngx_pool_t *pool_;
+
+public:
+  explicit NgxHeaderWriter(ngx_http_request_t *request)
+      : request_(request), pool_(request_->pool) {}
+
+  void set(std::string_view key, std::string_view value) override {
+    // Question: Should we check first if the key already exist?
```
dgoffredo, line 25: This question is worth revisiting.

dmehala:
Soon™️

PR #57, opened 2023-12-08
Remove NGINX_PATH_MODULES and NGINX_CONFIG_PATH

dmehala:
# Description 
The removal of `NGINXMODULES_PATH` and `NGINX_CONF_PATH` has the advantage of facilitating the execution of our tests, optimizing the utilization of matrix jobs on CircleCI, and streamlining the development workflow.

The goal here is to demonstrate a working solution and get some feedbacks.

### Changes
- Remove all occurrences of `NGINX_MODULES_PATH` and `NGINX_CONF_PATH`.
- Update `generate_jobs_yaml.sh` to leverage matrix jobs and regenerate `.circleci/config.yml`.
- Rework nginx service Dockerfile.
- Update the testing framework to align with new configuration path and module path.

## How does it works?
The module and its corresponding configuration are duplicated to a designated directory. Subsequently, the nginx is initiated utilizing the specified configuration file rather than the default configuration. I have updated the testing framework to:
- Substitute all instances of our module with its absolute location in the configuration.
- Modify the global environment `nginx_conf_path` to reflect the updated location of the configuration file.


dgoffredo (approved):
I added `NGINX_MODULES_PATH` and `NGINX_CONF_PATH` to support the testing of:

- debian flavored nginx
- alpine flavored nginx
- OpenResty (which has since been removed from testing)
- Amazon Linux

If you think that these changes obviate the need for these degrees of freedom, then we might as well do so.

I haven't tested your changes, but this looks good to me.

test/cases/orchestration.py
```diff
@@ -534,6 +543,10 @@ def nginx_test_config(self, nginx_conf_text, file_name):
         # -T            : test configuration, dump it and exit
         # -q            : suppress non-error messages during configuration testing
         # -c filename   : set configuration file (default: /etc/nginx/nginx.conf)
+        nginx_conf_text = nginx_conf_text.replace(
+            "load_module modules/ngx_http_datadog_module.so",
+            "load_module /test-conf/ngx_http_datadog_module.so",
+        )
```
dgoffredo, line 549: Per our discussion Thursday, please instead modify the actual test config files -- something like this:
```nginx
# "/test-conf" is a directory created by the docker build
# of the nginx test image. It contains the module, the
# nginx config, and "index.html".
load_module /test-conf/ngx_http_datadog_module.so;
```

dgoffredo (approved):
Thanks for this. It's a nice simplification.

Please make sure that this is consistent with the most recent `main` (some changes in `nginx-version-info`, etc.).

Also, as we discussed, I'd like for nginx-datadog to remain `make format` formatted, but it's likely that the `main` branch is not formatted so... either format this before merging, or we can deal with it later.

ship it squirrel

PR #59, opened 2024-01-24
feat: Set integration informations

dmehala:


dgoffredo (approved):
Looks good, thanks.

I ran the formatter as well. I see that you have a proposal to automate checking that. Another good idea.

PR #60, opened 2024-01-24
chore: enforce codestyle

dmehala:


dgoffredo (approved):
Thanks! I've committed some updates, so please have a look.

We eventually might want to revisit whether we can do better than a `Makefile` in this repository. For now, let's continue to use it.

The format CI check is failing because the dd-trace-cpp CI image that you've reused doesn't have `yapf3` in it. I think it's reasonable to add that to the `apt-get install` in dd-trace-cpp's `Dockerfile` and `docker push` it to the correct place. The ubuntu package is called `yapf3`. While we're at it, we can add Python formatting for the one `.py` file in dd-trace-cpp, or not.

Alternatively, if you have another approach, we can do that.

PR #78, opened 2024-05-15
Fix .circleci/config.yml

cataphract:
Yaml is a confusing format. To wit:

```
$ cat a.yml
a: "foobar"
b: foo "bar"
c: \"foobar\"
d: "foo\"bar"
$ ruby -e 'require "yaml"; a = YAML.load_file "a.yml"; puts a["a"]; puts a["b"]; puts a["c"]; puts a["d"]'
foobar
foo "bar"
\"foobar\"
foo"bar
```

end result was error:

```
datadog_version_ngx_mod mismatch: "1.2.0" != 1.2.0

Exited with code exit status 1
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ jobs:
         name: Verify versions (release tag only)
         command: |
           if [[ $CIRCLE_TAG =~ ^v ]]; then
-            bin/verify_version.sh "$(echo \"$CIRCLE_TAG\" | tr -d v)"
+            bin/verify_version.sh "$(echo "$CIRCLE_TAG" | tr -d v)"
           else
             echo "Not a release"
           fi


dmehala (approved):
YAML is a nutshell
image

Anilm3 (approved):

dgoffredo:
[Shell][1] is a confusing format.  Double quotes aren't escaped within a double-quoted command substitution.
```console
$ echo "how $(echo deep "$(echo we "$(echo can "$(echo go)")")")"
how deep we can go
```
I don't think that YAML was involved.

[1]: https://pubs.opengroup.org/onlinepubs/009604499/utilities/xcu_chap02.html#tag_02_02_03

smola (approved):

PR #82, opened 2024-05-20
Coverage with codecov.io

cataphract:
Generate coverage reports. See codecov comment

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,7 @@
 ---
 version: 2.1
+orbs:
+  codecov: codecov/codecov@4.1.0
 release_tag_only: &release_tag_only
   filters:
     tags:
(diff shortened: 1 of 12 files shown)


src/coverage_fixup.c
```diff
@@ -0,0 +1,50 @@
...
+  // https://github.com/llvm/llvm-project/blob/82c5d350d200ccc5365d40eac187b9ec967af727/compiler-rt/lib/profile/InstrProfilingFile.c#L870
+  char *modified_profile_file = malloc(strlen(profile_file) + 3);
+  if (profile_file[0] == '/') {
+    modified_profile_file[0] = '/';
+    strcpy(&modified_profile_file[1], profile_file);
+  } else {
+    modified_profile_file[0] = '.';
+    modified_profile_file[1] = '/';
+    strcpy(&modified_profile_file[2], profile_file);
+  }
```
Anilm3, line 37: I commend you for how hacky this is.

Anilm3 (approved):
Nothing much to say,  other than perhaps you could filter `coverage_fixup.c` and `glibc_compat.c` to avoid skewing the coverage (even if it's just a few %), you should be able to use `--ignore-filename-regex=` when calling llvm-cov.

dmehala (approved):

.circleci/config.yml
```diff
@@ -107,6 +109,27 @@ jobs:
...
+    - run:
+        command: 'make coverage'
+        environment:
+          ARCH: x86_64
+          MAKE_JOB_COUNT: 8
+          BUILD_TYPE: RelWithDebInfo
+          NGINX_VERSION: 1.26.0
+          WAF: ON
+    - codecov/upload:
+        file: .musl-build/coverage.lcov
```
dmehala, line 128: Even if you specified as file, codecovecli by default still search for coverage file. You might also want `upload_args: --disable-search`

PR #84, opened 2024-05-30
ci: Add system-tests

cbeauchesne:
Add a job in circleCI that run system-tests. It will prevent introducing bug tested in system tests.

For now, I only added the default scenario, up to C++ team to decide what to add or remove later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -161,6 +161,30 @@ jobs:
     - store_artifacts:
         path: test/logs/test.log
         destination: test.log
+  system_tests:
+    machine:
+      # https://support.circleci.com/hc/en-us/articles/360007324514-How-can-I-use-Docker-volume-mounting-on-CircleCI-
+      image: ubuntu-2004:current
+    resource_class: large
+    steps:
+      - attach_workspace:
+          at: "/tmp/workspace"
+      - run:
+          name: clone system-tests repo
+          command: git clone https://github.com/DataDog/system-tests.git
+      - run:
+          name: ls
+          command: ls -la /tmp/workspace/.musl-build
+      - run:
+          name: Build test targets
+          working_directory: ./system-tests
+          command: ./build.sh cpp
+      - run:
+          name: Run DEFAULT scenarios
+          working_directory: ./system-tests
+          command: ./run.sh
+
+
   format:
     docker:
     - image: datadog/docker-library:dd-trace-cpp-ci
@@ -285,6 +309,10 @@ workflows:
         filters:
           tags:
             ignore: "/^v[0-9]+\\.[0-9]+\\.[0-9]+/"
+    - system_tests:
+        name: Run system tests
+        requires:
+        - build 1.25.4 on amd64 WAF ON
   build-and-test-all:
     jobs:
     # output of bin/generate_jobs_yaml.rb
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -626,3 +658,7 @@ workflows:
         name: test << matrix.nginx-version >> on << matrix.base-image >>:<< matrix.arch >> WAF << matrix.waf >>
         requires:
         - build << matrix.nginx-version >> on << matrix.arch >> WAF << matrix.waf >>
+    - system_tests:
+        name: Run system tests
+        requires:
+        - build 1.25.4 on amd64 WAF ON
```
dmehala, line 664: This one is not necessary.

dmehala:
LGTM. However I can see in the system-tests output:

> Library: cpp@6.6.6
> Weblog variant: nginx

How the weblog versioning is working? Is there a weblog versioning?

dmehala (approved):

cbeauchesne:
aha, good catch, I didn't expect you to see this. There are several issue here.

Currently, the version number for c++ is the version number of this [present repo](https://github.com/DataDog/nginx-datadog/releases), where it should be the version of [dd-trace-cpp](https://github.com/DataDog/dd-trace-cpp/releases)

Second, I need to have a way to extract this version number from the `ngx_http_datadog_module.so` file, and I don't know how to do so. So shame on me, I went lazy, and I hardcoded a big version number, leaving this point for later.

If you know how to extract this version number, I can add it (it's inside system-tests repo). If you don't know, or if it's not possible (yet), my opinion is we can move forward with that hack, as long as the hardocded version is bigger than the current version, everything will works fine.

.circleci/config.yml
cbeauchesne: 🤔 it's necessay for `build-and-test` but not for `build-and-test-all` ?

cataphract:
> aha, good catch, I didn't expect you to see this. There are several issue here.
> 
> Currently, the version number for c++ is the version number of this [present repo](https://github.com/DataDog/nginx-datadog/releases), where it should be the version of [dd-trace-cpp](https://github.com/DataDog/dd-trace-cpp/releases)
> 
> Second, I need to have a way to extract this version number from the `ngx_http_datadog_module.so` file, and I don't know how to do so. So shame on me, I went lazy, and I hardcoded a big version number, leaving this point for later.
> 

They can be fetched like [this](https://github.com/DataDog/system-tests/pull/2455/files#diff-172cf498010a2e06a0e2e094374c274c96584c1eb6400c729a70be25ed0b17acR7).


See also: https://github.com/DataDog/nginx-datadog/blob/master/src/version.cpp.in

cbeauchesne:
thanks @cataphract , let me add this

dmehala:
> They can be fetched like [this](https://github.com/DataDog/system-tests/pull/2455/files#diff-172cf498010a2e06a0e2e094374c274c96584c1eb6400c729a70be25ed0b17acR7).

That's smart. We should also those this for `dd-trace-cpp` instead of parsing `version.h`. This would have avoid the issue with prod we got this week

cbeauchesne:
@dmehala could you merge the Pr? I don't have the right to do it myself 🙏

PR #88, opened 2024-06-04
fix: propagation header value adding up

dmehala:
This commit fix a regression introduced in #79.
During the context propagation phase, if an HTTP header already exist, the value is appended instead of being overwritten.

Changes:
  - Add integration test covering the issue.
  - Update header injection to overwrite the value for an existing header name.


pablomartinezbernardo (approved):
LGTM

PR #93, opened 2024-06-25
feat: add RUM SDK Injection

dmehala:
# Description
First implementation of a filter module to automatically inject the RUM Browser SDK.


requirements.txt
```diff
@@ -0,0 +1,2 @@
+werkzeug==2.2.3
```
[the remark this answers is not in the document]
dmehala: TODO: Upgrade to `3.0.3`

CMakeLists.txt
```diff
@@ -156,9 +176,10 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   set_target_properties(ngx_http_datadog_module PROPERTIES LINK_DEPENDS ${LINKER_SCRIPT})
   target_link_options(ngx_http_datadog_module PRIVATE "-Wl,--version-script=${LINKER_SCRIPT}")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-  set(EXPORTED_SYMBOLS ${CMAKE_SOURCE_DIR}/nginx_module.exported)
-  set_target_properties(ngx_http_datadog_module PROPERTIES LINK_DEPENDS ${EXPORTED_SYMBOLS})
-  target_link_options(ngx_http_datadog_module PRIVATE "-Wl,-exported_symbols_list" "-Wl,${EXPORTED_SYMBOLS}")
+  # set(EXPORTED_SYMBOLS ${CMAKE_SOURCE_DIR}/nginx_module.ld)
+  # set_target_properties(ngx_http_datadog_module PROPERTIES LINK_DEPENDS ${EXPORTED_SYMBOLS})
+  # target_link_options(ngx_http_datadog_module PRIVATE "-Wl,-exported_symbols_list" "-Wl,${EXPORTED_SYMBOLS}")
+  target_link_options(ngx_http_datadog_module PRIVATE -undefined dynamic_lookup)
```
dmehala, line 182: - [ ] revert

test/cases/rum/test_injection.py
```diff
@@ -0,0 +1,377 @@
...
+window.DD_RUM.onReady(function() {
+  window.DD_RUM.init({"applicationId":"<DATADOG_APPLICATION_ID>","clientToken":"<DATADOG_CLIENT_TOKEN>","env":"production","service":"my-web-application","sessionReplaySampleRate":100.0,"sessionSampleRate":100.0,"site":"<DATADOG_SITE>","trackLongTasks":true,"trackResources":true,"trackUserInteractions":true,"version":"1.0.0"});
+});
+</script>
+</head>
+<body>
+
+<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Dolor magna eget est lorem ipsum. Vulputate sapien nec sagittis aliquam malesuada bibendum arcu vitae. Tellus mauris a diam maecenas sed enim ut sem. Urna duis convallis convallis tellus id. Gravida neque convallis a cras semper auctor neque vitae. Enim neque volutpat ac tincidunt. Commodo viverra maecenas accumsan lacus. Egestas sed tempus urna et pharetra pharetra massa massa. Ac auctor augue mauris augue neque. Ut placerat orci nulla pellentesque dignissim enim sit amet venenatis. Ultrices mi tempus imperdiet nulla malesuada pellentesque elit. Egestas diam in arcu cursus euismod. Consequat semper viverra nam libero justo.
```
pablomartinezbernardo, line 35: Do we need the full lorem ipsum? 😅

src/rum/injection.cpp
```diff
@@ -0,0 +1,257 @@
...
+  if (h == nullptr) {
+    state_ = state::error;
+    ngx_log_error(NGX_LOG_ERR, r->connection->log, 0,
+                  "RUM SDK injection failed: unable to add "
+                  "x-datadog-sdk-injected HTTP header");
+    return NGX_ERROR;
+  }
+
+  h->hash = 1;
+  ngx_str_set(&h->key, "x-datadog-sdk-injected");
```
pablomartinezbernardo, line 141: What is different in behavior beween `x-datadog-sdk-injected` and `x-datadog-rum-injected`? Should we be checking for the same headers and not use two?

src/rum/injection.cpp
```diff
@@ -0,0 +1,257 @@
...
+    return &h[i];
+  }
+
+  return nullptr;
+}
+
+bool is_html_content(ngx_str_t *content_type) {
+  assert(content_type != nullptr);
+  std::string_view content_type_sv = to_string_view(*content_type);
+  return content_type_sv.find("text/html") != content_type_sv.npos;
```
pablomartinezbernardo, line 48: Should we inject on `application/xhtml+xml` as well?

src/rum/config.cpp
```diff
@@ -0,0 +1,154 @@
...
+  char *status = ngx_conf_parse(cf, NULL);
+  *cf = save;
+
+  if (status != NGX_CONF_OK) {
+    return status;
+  }
+
+  auto json = make_rum_json_config(config_version, rum_config);
+  if (json.empty()) {
+    auto *err_msg =
```
pablomartinezbernardo, line 122: Forgot to return `err_msg` here?

src/rum/config.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+static std::string make_rum_json_config(
+    std::string_view config_version,
+    const std::unordered_map<std::string, std::string> &config) {
+  rapidjson::Document doc;
+  doc.SetObject();
+
+  rapidjson::Document::AllocatorType &allocator = doc.GetAllocator();
+  doc.AddMember("majorVersion",
+                rapidjson::Value(std::stoi(config_version.data() + 1)),
```
pablomartinezbernardo, line 54: Edge case, but I think this will break if the config_version is just `v`

test/cases/rum/test_injection.py
dmehala: Yes.

src/rum/injection.cpp
dmehala: we should check only for `x-datadog-rum-injected`. The other one is from an older time where we didn't fully agree on the name.

src/rum/injection.cpp
dmehala: `xhtml` is a stricter form of html. The browser parse and interpret it as XML. The content inside the `script` tag might be interpreted as XML, to prevent, the JS code must be wrapped in a `CDATA` section. At the moment the injected script is not compliant with xhtml, meaning, we should not inject on xhtml.

src/rum/config.cpp
dmehala: Indeed. Good catch.
- [ ] return `err_msg`

src/rum/config.cpp
dmehala: This will definitely throw an uncatch exception. Good catch (no pun intended).
- [ ] sanitize args input.

test/cases/rum/test_injection.py
```diff
@@ -0,0 +1,377 @@
...
+            - Append `x-datadog-sdk-*` headers
+        """
+        status, lines = self.load_conf("rum_enabled.conf")
+        self.assertEqual(0, status, lines)
+
+        status, headers, body = self.orch.send_nginx_http_request(
+            "/dd_logo_v_rgb.png")
+        headers = self.make_dict_headers(headers)
+        self.assertEqual(status, 200)
+        self.assertTrue("x-datadog-sdk-injected" not in headers)
```
pablomartinezbernardo, line 354:
```suggestion
        self.assertTrue("x-datadog-rum-injected" not in headers)
```

test/cases/rum/test_injection.py
```diff
@@ -0,0 +1,377 @@
...
+        turn off the injection.
+        """
+        status, lines = self.load_conf("rum_enabled.conf")
+        self.assertEqual(0, status, lines)
+
+        status, headers, body = self.orch.send_nginx_http_request(
+            "/disable-rum")
+        headers = self.make_dict_headers(headers)
+        assert status == 200
+        assert headers.get("x-datadog-sdk-injected") == None
```
pablomartinezbernardo, line 372:
```suggestion
        assert headers.get("x-datadog-rum-injected") == None
```

pablomartinezbernardo (approved):
Only some small changes in the tests

PR #95, opened 2024-06-28
fix: datadog script execution

### LINKED ISSUE #94: Error: Executing invalid Datadog script
After updating to the version `1.2.x` without appsec support we are seeing the following nginx errors in our logs:

```shell
2024/06/27 18:16:19 [error] 33#33: *2 Executing invalid Datadog script, client: 192.168.65.1, server: localhost, request: "GET / HTTP/1.1", host: "localhost:8080"
2024/06/27 18:16:19 [error] 33#33: *2 Executing invalid Datadog script, client: 192.168.65.1, server: localhost, request: "GET / HTTP/1.1", host: "localhost:8080"
2024/06/27 18:16:19 [error] 33#33: *2 Executing invalid Datadog script while logging request, client: 192.168.65.1, server: localhost, request: "GET / HTTP/1.1", host: "localhost:8080"
```

Logs and APM traces are published in the Datadog dashboard, but the fact that these errors occur for each request is a bit concerning.  Is there a new Datadog module setting we need to set?  

When we revert back to `v1.1.0` these errors are NOT present. 

Tech Specifications:
Alpine Linux amd64:`v3.20.0`
nginx: `v1.2.1`
nginx-datadog module: `v1.2.x`

dmehala:
## Description
This commit fixes a regression likely introduced in PR #79. The module attempts to resolve an uninitialized configuration variable, resulting in an error log.

Resolves #94.

## What is happening?
The sampling delegation feature introduced two directives: `datadog_delegate_sampling` and `datadog_allow_sampling_delegation_in_subrequests`.

These directives can accept variables and requires NGINX to resolve these variables while processing the request to determine if sampling should be delegated. Since #79, the default values for these directives have not been set, resulting in a malformed `NgxScript`.

## The solution
I removed the ability to extend variables for sampling delegation directives and ensured the underlygin variables are initialized.


nginx_module.ld-prime
```diff
@@ -0,0 +1,8 @@
+# nm -m to get the listo f symbols
```
dmehala, line 1:
```suggestion
# `nm -m` to get the list of symbols
```

PR #98, opened 2024-08-01
build: move glibc_compat to toolchain

cataphract:
Also do some fixes on the *stat functions (fixed the first argument) and add getrandom for old glibc versions (like xenial)


Anilm3 (approved):
lgtm

PR #101, opened 2024-08-03
feat: trace request handled by `ngx_http_rewrite_module`

### LINKED ISSUE #58: `return` directives not traced due to `ngx_module_order` declaration
# Description
Currently, there is an issue where `return` directives are not being traced. Generally speaking `ngx_http_rewrite_module` requests are not traced. It is a regression from the legacy opentracing module.

## Why?
Upon investigation, it has been identified that the problem is related to the declaration of `ngx_module_order` in [module/config](https://github.com/DataDog/nginx-datadog/blob/4745ac3bad5f3c7a924e1109069fb24fe708be45/module/config#L9). According to the nginx documentation, `ngx_module_order` is used to:

> Set the load order for the module which is useful for HTTP_FILTER and HTTP_AUX_FILTER module types.

> The order is stored in a reverse list. The ngx_http_copy_filter_module is near the bottom of the list so is one of the first to be executed. This reads the data for other filters. Near the top of the list is ngx_http_write_filter_module which writes the data out and is one of the last to be executed.

In our case, the ordering is necessary for allowing our module to hijack some directives. However, this creates a conflict, as changing the order does not only affects the initialisation order but also the execution order. As a result, requests using the `ngx_http_rewrite_module` are executed before our module and its never being called.

## Additional information
- About the order of execution of the modules: https://mailman.nginx.org/pipermail/nginx/2021-March/060514.html

dmehala:
## Description
This commit adds the ability to trace requests handled by the rewrite module. Previously, configuration like the one in the following example wouldn't be traced, even through they would be if using [nginx-opentracing](https://github.com/DataDog/nginx-opentracing).

```
...
location /foo {
   return 200 "hello, world!";
}
...
```

This is prerequisites for PR #93.
Resolves #58.

## Motivation
By default, the module automatically appends Trace ID and Span ID to access logs to provide a seamless experience with out-of-the-box [trace and log correlation](https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/).

To achieve this seamless experience, several workarounds were used. One key workaround involved loading the module before the log module (`ngx_http_log_module`) to hijack the `access_log` directive and modify its behaviour. 

While this approach offered a smooth user experience, it led to inconsistencies and regressions in tracing certain requests. It also prevent use to define response filter required to support RUM SDK Injection. 

## How Does It Work?
An NGINX module's loading order can be controlled using `ngx_module_order`. The execution order is the reverse of the loading order. For instance, `ngx_module_order=foo_module bar_module` can be interpreted as:
* loading order: `foo_module` then `bar_module`.
* execution order: `bar_module` then `foo_bar`.

For more details, refer to this [NGINX mailing list discussion](https://mailman.nginx.org/pipermail/nginx/2021-March/060514.html).

To hijack the `access_log` directive, the module needs to be loaded before `ngx_http_log_module`. However rewrite module (`ngx_http_rewrite_module`) depends on the logging module, our was then loaded before the rewrite module and thus executed after it.

Consequently, this workaround has to be removed and log - trace correlation are not an out-of-the-box experience anymore 😕.


pablomartinezbernardo (commented):
How would a user set up log injection after this?

dmehala:
By defining themselves the format:
```
log_format datadog_text '$remote_addr - $http_x_forwarded_user [$time_local] "$request" $status $body_bytes_sent "$http_referer" "$http_user_agent" "$http_x_forwarded_for" "$datadog_trace_id" "$datadog_span_id"';

log_format datadog_json escape=json '{"remote_addr": "$remote_addr", "forwarded_user": "$http_x_forwarded_user", "time_local": "$time_local", "request": "$request", "status": $status, "body_bytes_sent": $body_bytes_sent, "referer": "$http_referer", "user_agent": "$http_user_agent", "forwarded_for": "$http_x_forwarded_for", "dd.trace_id": "$datadog_trace_id", "dd.span_id": "$datadog_span_id"}';
```

I plan to update the public documentation.

pablomartinezbernardo (approved):
Cool, defining a custom log format doesn't seem that terrible, it's worth the trade off

PR #103, opened 2024-08-09
feat: use hex representation for `$datadog_trace_id` and `$datadog_parent_id` variables

### LINKED ISSUE #102: Allow otel trace ids to be output in logs
Reading the [variables section of the docs](https://github.com/DataDog/nginx-datadog/blob/master/doc/API.md#variables), we can access `datadog_trace_id` and `datadog_span_id` which are integers. It would be great to be able to instead output the OpenTelemetry version of trace and span ids based on a hex string instead.

This would be similar to how `traceparent` works just with the ability to write those variables to logs.

Thanks!

dmehala:
Changes:
- Add two variables: $datadog_trace_id_dec and $datadog_span_id_dec.
- Update API.md with those two variables.
- Warn at startup about the change of behaviour for $datadog_trace_id and $datadog_span_id.

Resolves #102

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -476,6 +476,14 @@ Note that if `datadog_trace_locations` is `on`, then `$datadog_span_id` will
 refer to the span associated with the location (outbound request), not its
 parent (inbound request).
 
+### `datadog_trace_id_dec`
+Same as `$datadog_trace_id`, but in decimal format. Up until v1.2.1, this was
+the value returned by `$datadog_trace_id`.
+
+### `datadog_span_id_dec`
+Same as `$datadog_span_id`, but in decimal format. Up until v1.2.1, this was
+the value returned by `$datadog_span_id`.
+
 ### `datadog_location`
 `$datadog_location` expands to the name or pattern of the `location` block that
 matched the current request.  For example,
--- a/doc/API.md
+++ b/doc/API.md
@@ -464,25 +464,23 @@ about the currently active trace.
 
 ### `datadog_trace_id`
 `$datadog_trace_id` expands to the decimal representation of the unsigned
-64-bit ID of the currently active trace.  If there is no currently active
+64-bit ID of the currently active trace. If there is no currently active
 trace, then the variable expands to a hyphen character (`-`) instead.
 
 ### `datadog_span_id`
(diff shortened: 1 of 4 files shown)


dgoffredo:
Flyby comment. Maybe I should disable notifications about these repositories. :)

This change could cause problems for users who depend on those values currently and then upgrade.

Consider adding new variables for the hex representation. You probably already thought of that.

Maximal solution:
1. `$datadog_trace_id_hex` is the 128-bit trace ID in hexadecimal.
2. `$datadog_span_id_hex` is the 64-bit span ID in hexadecimal.
3. `$datadog_trace_id_dec` is the 64-bit (!) trace ID in decimal.
4. `$datadog_span_id_dec` is the 64-bit trace ID in decimal.
5. `$datadog_trace_id` is `$datadog_trace_id_dec`.
6. `$datadog_span_id` is `$datadog_span_id_dec`.

The tradeoff is between "backward compatibility" and "OpenTelemetry is the way." If 128-bit hex IDs really are where Datadog is going, then consider the above but with (5) and (6) changed to prefer hex, so that then you can print a big warning on the next release: "Trace ID and span ID variables are now hexadecimal! Use $foo_dec for the old behavior!"

dmehala:
Hey @dgoffredo ! Thanks for chiming in.

I did considered new variables, but the more I reflect on it, the less. certain I am that we actually need them. My thinking is `$datadog_X_id` doesn't serve much prupose for customers beyond log correlation or logging - though I could be wrong.

In the past, when I made a breaking change without notifying our customers, we received some complaints. So, I can almost hear Caleb saying something along the line "you should warn first any future breaking change". Since you're suggesting the same, it's probably wise to follow that approach.

Thanks again!

pablomartinezbernardo (commented):
I'd like to push back a bit on using the hex value by default. I agree with the assumption "OTel is the future", but I am also assuming customers who will be annoyed by this change are more right now than those who will benefit from hex being the default.

dmehala:
Hm, I see your point.

I will define `$datadog_x_hex` and log a warning that the default behaviour of `$datadog_x` will change in the N+1 release.
Thank you for your feedback @pablomartinezbernardo .

phillipuniverse:
Just got around to testing this and looks like it's not working, opened a bug at #166.

PR #105, opened 2024-08-12
Override pthread_cond_init in glibc-compat

Anilm3:
`pthread_cond_init` in glibc is provided in two different versions: the "current" version mapped to `__pthread_cond_init`, and the legacy version mapped to `__pthread_cond_init_2_0`. When using this function through a musl-built binary, the dynamic linker opts to resolve the non-default legacy version, presumably for backwards compatibility. 

However, the legacy API for condition variables forbids the use of clocks other than `CLOCK_REALTIME`, while in many cases `CLOCK_MONOTONIC` is preferred instead, resulting in an unexpected failure to initialise a condition variable.

The fix provided overrides `pthread_cond_init` with a weak symbol and then resolves the default `pthread_cond_init` provided in `libc.so.6`. Note that this is only a problem in `x86_64`.

The images will be updated once this has been merged.

### DIFF
--- a/build_env/glibc_compat.c
+++ b/build_env/glibc_compat.c
@@ -132,6 +132,53 @@ int pthread_atfork(
     return real_atfork(prepare, parent, child);
 }
 
+struct pthread_cond;
+struct pthread_condattr;
+typedef struct pthread_cond pthread_cond_t;
+typedef struct ipthread_condattr pthread_condattr_t;
+
+int __pthread_cond_init (pthread_cond_t *cond, const pthread_condattr_t * attr) __attribute__((weak));
+
+int pthread_cond_init(pthread_cond_t *cond, const pthread_condattr_t *cond_attr)
+{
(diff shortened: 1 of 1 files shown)


dmehala (approved):
LGTM! :shipit: 
How did you check if it does behave as intended?

Anilm3:
> LGTM! :shipit: How did you check if it does behave as intended?

Essentially validating that this works as expected, i.e. `error == 0`. When using the old symbol, `pthread_cond_init` results in errno being set to `EINVAL (22)`.

```
pthread_condattr_init(&attr);
pthread_condattr_setclock(&attr, CLOCK_MONOTONIC);
error = pthread_cond_init(&cond_, &attr);
```

But this has also been indirectly validated through the PHP appsec CI and system tests, which were failing due to the remote configuration client failing to initialise, specifically due to boost `io_context` not being able to initialise the condition variable.

PR #106, opened 2024-08-14
Track new `dd-trace-cpp` API usage

dmehala:
Issues:
- [ ] Test failing due to missing JSON conf? or ill-formatted?


pablomartinezbernardo (approved):
I'm failing to see why we need to skip that test, what has changed that broke the behavior in the test?

Otherwise LGTM

PR #109, opened 2024-08-27
chore(release): Use tarfile to package the module

dmehala:
# Description
Resolves warning message during the extraction of the module distributed on releases.


bin/release.py
```diff
@@ -159,35 +170,19 @@ def prepare_release_artifact(work_dir, build_job_number, version, arch, waf):
...
-        work_dir,
-        module_debug_path.name,
-    ]
-    run(command, check=True)
-
-    command = [gpg_exe, "--armor", "--detach-sign", debug_tarball_path]
-    run(command, check=True)
-
-    module_path.unlink()
-    module_debug_path.unlink()
```
pablomartinezbernardo, line 190: Is there any reason to not delete the files now that did apply before this change? Or were these removed because they are not necessary?

pablomartinezbernardo (approved):
LGTM

PR #110, opened 2024-09-03
Initial installer version

pablomartinezbernardo:
TODO for a future PR:

- [x] CI: Build installer and publish in releases, along with script
- [x] Actually download binary
- [x] Verify signatures
- [ ] Telemetry on both the script and installer


rum-injection-installer/installer/installer.go
```diff
@@ -0,0 +1,67 @@
...
+func validateInput(appID, site, clientToken, arch, agentUrl string, sessionSampleRate, sessionReplaySampleRate int) error {
+	if appID == "" || site == "" || clientToken == "" || agentUrl == "" {
+		return fmt.Errorf("appID, site, clientToken, and agentUrl are required")
+	}
+	if sessionSampleRate < 0 || sessionSampleRate > 100 {
+		return fmt.Errorf("sessionSampleRate must be between 0 and 100")
+	}
+	if sessionReplaySampleRate < 0 || sessionReplaySampleRate > 100 {
+		return fmt.Errorf("sessionReplaySampleRate must be between 0 and 100")
+	}
```
dmehala, line 18: Those fields are validated by the NGINX module when `nginx -t` is executed. I believe it is best to delegate the validation. What do you think?

rum-injection-installer/installer/installer.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if err := configurator.ModifyConfig(*appID, *site, *clientToken, *agentUrl, *sessionSampleRate, *sessionReplaySampleRate); err != nil {
+		handleError(err)
+	}
+
+	if err := configurator.ValidateConfig(); err != nil {
+		handleError(err)
+	}
+
+	fmt.Println("Configuration completed successfully")
```
dmehala, line 66:
```suggestion
	fmt.Println("Datadog NGINX module has been successfully installed and configured. Please restart NGINX for the changes to take effect")
```

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+type NginxConfigurator struct {
+	Version     string
+	ModulesPath string
+}
+
+// Runs 'nginx -v' and looks for the nginx version.
+// Checks that the version is higher than a min version
+func (n *NginxConfigurator) Verify() error {
```
dmehala, line 21: **[suggestion]** the function name is a bit too vague. What is being verified? May I suggest `VerifyRequirement`?

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	cmd := exec.Command("nginx", "-v")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("nginx is not installed")
```
dmehala, line 25:
```suggestion
		return fmt.Errorf("nginx was not detected. Please ensure NGINX is included in the PATH")
```

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	matches := versionRegex.FindStringSubmatch(string(output))
+	if len(matches) < 2 {
+		return fmt.Errorf("failed to parse nginx version: %s", output)
+	}
+
+	n.Version = matches[1]
+
+	print("Detected NGINX version: ", n.Version, "\n")
+
+	const minSupportedVersion = "1.22.0"
```
dmehala, line 38: The minimum version supported will change over time. The best way to know the minimum supported version is to get the list of module in the latest release.  You can retrieve this list using the Github API.

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	return io.ReadAll(resp.Body)
+}
+
+// Gets the configuration file from 'nginx -T'
+func (n *NginxConfigurator) getConfigLocation() (string, error) {
```
dmehala, line 208:
```suggestion
func (n *NginxConfigurator) getDefaultConfigLocation() (string, error) {
```

It reflect better what the function does.

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+
+	content, err := os.ReadFile(configPath)
+	if err != nil {
+		return fmt.Errorf("failed to read config file: %v", err)
+	}
+
+	// Check if Datadog module is already loaded
+	if strings.Contains(string(content), "ngx_http_datadog_module.so") {
+		return fmt.Errorf("datadog module is already loaded in the configuration")
+	}
```
dmehala, line 65: The following may return a false positive if the Datadog module is named differently. To ensure the module isn't already loaded, you can attempt to load it—NGINX will return an error if it is already installed.

```sh
# the following will fail if the default config already load the datadog module
nginx -g "load_module <MODULE>;" -t

nginx: [emerg] module "ngx_http_datadog_module" is already loaded in /Users/damien.mehala/workspace/nginx-datadog/nginx.conf:2
```

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+
+	var newContent strings.Builder
+	newContent.WriteString(fmt.Sprintf("load_module %s/ngx_http_datadog_module.so;\n", n.ModulesPath))
+
+	// Find http and bracket { on possibly multiple lines
+	httpRegex := regexp.MustCompile(`(?m)^[\s]*http[\s]*\{`)
+
+	httpBlockStart := httpRegex.FindIndex(content)
+	if httpBlockStart == nil {
+		return fmt.Errorf("could not find http block in nginx config")
```
dmehala, line 82: **[suggestion]** include the filepath

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+
+	newContent.Write(content[:httpBlockStart[0]])
+
+	newContent.Write(content[httpBlockStart[0]:httpBlockStart[1]])
+	newContent.WriteString("\n")
+
+	datadogConfig := fmt.Sprintf(`    datadog_agent_url %s;
+    datadog_disable;
+    datadog_rum on;
```
dmehala, line 92:
```suggestion
    # Disable APM Tracing. Remove the next line to enable APM Tracing.
    datadog_disable;
  
    # Enable RUM Injection
    datadog_rum on;
```

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+        "site" "%s";
+        "sessionSampleRate" %d;
+        "sessionReplaySampleRate" %d;
+    }
+`, agentUrl, appID, clientToken, site, sessionSampleRate, sessionReplaySampleRate)
+	newContent.WriteString(datadogConfig)
+
+	newContent.Write(content[httpBlockStart[1]:])
+
+	// Write the modified content back to the file
```
dmehala, line 105: I would feel more confortable to write the modified configuration somewhere on disk, validate it with `nginx -c <CONFIG> -t`, then, swap with the default config if valid, otherwise, abort and and delete the modified configuration.

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create temp directory: %v", err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	moduleFile := filepath.Join(tmpDir, "module.tgz")
+	signatureFile := filepath.Join(tmpDir, "module.tgz.asc")
+	publicKeyFile := filepath.Join(tmpDir, "pubkey.gpg")
+
+	if err := os.WriteFile(moduleFile, moduleContent, 0644); err != nil {
```
dmehala, line 160: **[question]** Mind explaining why `0644`? Is it required?

rum-injection-installer/installer/nginx_configurator.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return nil
+}
+
+// Downloads the latest nginx-datadog for the specified architecture
+// Extracts the .so file and places it in the modules path
+func (n *NginxConfigurator) DownloadAndInstallModule(arch string) error {
+	if err := n.getModulesPath(); err != nil {
+		return err
+	}
```
dmehala, line 129: It's not strictly required to place the module in the default module path. For instance, you can store it in `/datadog/ngx_http_datadog_module.so`, and as long as the `load_module` directive specifies the absolute path, it should function correctly.

dmehala (commented):
Thank you! Overall it is a good start, I left a few comments. I would like to decouple the installer from RUM, so customer could start using it to install APM tracing and/or appsec. What's your thought?

Here is an exhaustive list of missing features:
- dry run.
- verbose: being able to increase the verbosity to have a better understanding of what is going on or debug.
- tests for shell and go.
- versionning of the installer.

bin/release.py
```diff
@@ -142,6 +142,29 @@ def sign_package(package_path: str) -> None:
     run(command, check=True)
 
 
+def prepare_installer_release_artifact(work_dir, build_job_number, arch):
```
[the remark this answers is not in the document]
dmehala: Apologies for the confusion. The question isn't about the functionality of the script itself but rather how you plan to distribute the installer.

If I understood correctly, you intend to package and distribute the script with *each* release. Is that correct? If so, it means that for every release of the NGINX module, there would also be an installer, which could potentially create a release cycle dependency between the script and the module. 

Have you thought about this? Is it what you want? Is it a necessity to distribute the installer that way?

rum-injection-installer/install_script.sh
```diff
@@ -0,0 +1,112 @@
...
+            ;;
+    esac
+
+    echo "Detected architecture: $ARCH"
+}
+
+check_dependencies() {
+    command -v curl >/dev/null 2>&1 || error "curl is not installed. Please install curl and try again."
+    command -v gpg >/dev/null 2>&1 || error "gpg is not installed. Please install gpg and try again."
+    command -v tar >/dev/null 2>&1 || error "tar is not installed. Please install tar and try again."
```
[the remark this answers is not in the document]
pablomartinezbernardo: > I propose the new wording. It has the advantage to explain the customer why curl, gpg and tar is required.
> [question] I don't believe gpg should be an hard dependency. The customer should have the possibility to skip the verification.a

👍 will do

> [suggestion] do Go has a tar library, similar to [tarfile](https://docs.python.org/3/library/tarfile.html)?

The go code unpackages using the [official gzip package](https://pkg.go.dev/compress/gzip). But we also need to extract the installer itself, so it needs to be used from the script.

rum-injection-installer/install_script.sh
```diff
@@ -0,0 +1,112 @@
...
+    BASE_URL="https://github.com/DataDog/nginx-datadog/releases/latest/download"
+    BINARY_URL="${BASE_URL}/nginx-configurator-${ARCH}.tgz"
+    SIGNATURE_URL="${BINARY_URL}.asc"
+    PUBKEY_URL="${BASE_URL}/pubkey.gpg"
+
+    curl -sSLO "$BINARY_URL" || error "Failed to download binary"
+    curl -sSLO "$SIGNATURE_URL" || error "Failed to download signature"
+    curl -sSLO "$PUBKEY_URL" || error "Failed to download public key"
+
+    gpg --import pubkey.gpg || error "Failed to import public key"
```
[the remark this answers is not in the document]
pablomartinezbernardo: Unluckily it is for ascii-armored format, which the key is

rum-injection-installer/installer/installer.go
pablomartinezbernardo: Sort of disagree, I think making sure (As much as possible) that the installer doesn't produce a wrong output is a good idea. Otherwise, I don't see much issue with repeating the validation, and do see the value in being able to go to a single place to view/update parameter requirements.

dmehala: > I think making sure (As much as possible) that the installer doesn't produce a wrong output is a good idea.

I agree, that's why the installer must run `nginx -t` to validate the configuration before "committing" the new configuration file. Since NGINX is the one making the call, why not defer the whole validation to NGINX? 

The benefit is not just repeating validation but also avoid the maintenance burden.

rum-injection-installer/installer/nginx_configurator.go
pablomartinezbernardo: It is required for this API, which writes to file in a single call. There's another API that would let us not specify permissions, instead uses a default 666 for all files created,but we need to do more error handling, defer closing the files... Another thing to mention is `WriteFile` doesn't change permissions if the file already exists, so we are not messing with their actual config file. 644 is just the safest permissions I could think about that wouldn't cause problems, but I'm happy with 666 as well if you'd rather have that

rum-injection-installer/installer/nginx_configurator.go
pablomartinezbernardo: Oh I didn't know that thanks, this was the part of the code I trusted the least, so it makes my life easier

.circleci/config.yml
```diff
@@ -108,6 +108,34 @@ jobs:
       MAKE_JOB_COUNT: 8
       WAF: "<< parameters.waf >>"
       NGINX_VERSION: "<< parameters.nginx-version >>"
+  build_arm64_installer:
```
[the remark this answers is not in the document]
pablomartinezbernardo: > dry run.
> verbose.

👍 

> tests for shell and go.

WDYT of leaving this for later?

> versioning

I guess this depends on [this other conversation](https://github.com/DataDog/nginx-datadog/pull/110#discussion_r1745602587)

> download and verify the go binary.

Can you clarify what you are referring to with this?

> To get data we can get this installer in the wild for tracing and appsec but that requires a why to select the feature needed.

Do you mean letting the customer choose WAF ON/OFF for example? Do you think we should be doing that now given the configuration will be completely correlated to rum injection?

rum-injection-installer/installer/installer.go
pablomartinezbernardo: Still disagree, I don't think the "maintenance burden" is relevant for some strings and integers in 6 lines of code, and lets the user realize this is an error in the arguments early in the validation rather than an error with the installer that was unable to produce a good output

bin/release.py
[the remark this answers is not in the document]
dmehala: My fear is one needs to use a specific version of the installer to install the specified version of the NGINX module. For example, one must use v1.2.1 installer to install v1.2.1 of the module.

IMO, a version agnostic installer offers better value. By that, I mean, an installer being able to install any versions of the NGINX module. Such agnostic installer do not need to be released for each version of the module. What do you think?

bin/release.py
```diff
@@ -142,6 +142,29 @@ def sign_package(package_path: str) -> None:
...
+    artifacts = send_ci_request_paged(
+        f"/project/{PROJECT_SLUG}/{build_job_number}/artifacts")
+    module_url = None
+    for artifact in artifacts:
+        name = artifact["path"]
+        if name == "nginx-configurator":
+            module_url = artifact["url"]
+
+    if module_url is None:
+        raise Exception(
```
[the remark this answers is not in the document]
dmehala: That's a very opinionated 😆

.circleci/config.yml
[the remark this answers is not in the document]
dmehala: > Do you mean letting the customer choose WAF ON/OFF for example? 
Yes, you're absolutely right. WAF requires changes to the configuration, so it might not be the best time to introduce that. However, for APM tracing, only the `load_module <MODULE_PATH>;` is needed. If a customer is using the installer specifically for APM tracing, the current code should be sufficient, right?

installer/configurator/nginx_configurator.go
```diff
@@ -160,22 +136,120 @@ func (n *NginxConfigurator) DownloadAndInstallModule(arch string) error {
...
-		return fmt.Errorf("failed to write public key file: %v", err)
-	}
 
-	if err := verifySignatureGPG(moduleFile, signatureFile, publicKeyFile); err != nil {
-		return fmt.Errorf("failed to verify module signature: %v", err)
+	if !skipVerify {
+		signatureURL := moduleURL + ".asc"
+		signatureContent, err := downloadFile(signatureURL)
+		if err != nil {
+			return fmt.Errorf("failed to download signature: %v", err)
```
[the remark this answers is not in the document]
pablomartinezbernardo: I'm not following this one, what would you do with the other error messages for verification? Keep them with that same log?

installer/configurator/nginx_configurator.go
```diff
@@ -160,22 +136,120 @@ func (n *NginxConfigurator) DownloadAndInstallModule(arch string) error {
...
+	}
+
+	// Test again loading the module
+	cmd := exec.Command("nginx", "-g", fmt.Sprintf("load_module %s;", modulePath), "-t")
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		if strings.Contains(string(output), "already loaded") {
+			return fmt.Errorf("datadog module is already loaded, can't perform installation: %s\n%s", err, string(output))
+		}
+		return fmt.Errorf("nginx configuration test with module failed with unexpected error: %s\n%s", err, string(output))
```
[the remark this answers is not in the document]
pablomartinezbernardo: Sadly, `nginx -t` does not return error codes for specific errors, only 0 and 1

installer/configurator/nginx_configurator.go
```diff
@@ -222,51 +296,16 @@ func (n *NginxConfigurator) getConfigLocation() (string, error) {
...
-		if len(matches) > 1 {
-			path := matches[1]
-			if _, err := os.Stat(path); err == nil {
-				n.ModulesPath = path
-				return nil
-			}
-		}
-	}
+func (n *NginxConfigurator) createModulesDir() error {
+	modulesPath := "/datadog"
```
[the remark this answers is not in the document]
pablomartinezbernardo: From [the docs](https://pkg.go.dev/os#MkdirAll):

> If path is already a directory, MkdirAll does nothing and returns nil

installer/configurator/nginx_configurator.go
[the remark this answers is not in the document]
dmehala: In that case `!= 0` is enough right? Instead of looking for a string. By experience, relying on string is fragile.

.circleci/config.yml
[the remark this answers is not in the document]
pablomartinezbernardo: Should be pretty close, but the config changes as well, we are right now disabling apm with `datadog_disable` and forcing rum on. Are you suggesting adding a parameter determining whether `datadog_disable` & `datadog_rum` should be included or not?

installer/configurator/utils.go
```diff
@@ -0,0 +1,114 @@
...
+
+		dest := filepath.Join(destPath, header.Name)
+
+		switch header.Typeflag {
+		case tar.TypeReg:
+			f, err := os.OpenFile(dest, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
+			if err != nil {
+				return err
+			}
+			if _, err := io.Copy(f, tarReader); err != nil {
```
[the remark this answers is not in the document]
pablomartinezbernardo: Irrelevant due to our download's size

dmehala (approved):
LGTM :shipit:

installer/install-nginx-datadog.sh
```diff
@@ -0,0 +1,144 @@
...
+
+    echo "Binary downloaded, verified, and extracted successfully."
+}
+
+main() {
+    check_dependencies
+
+    ARGS_COPY=$*
+    SKIP_VERIFY=false
+    AGENT_URI="http://localhost:8126"
```
labbati, line 117: @pablomartinezbernardo  Should we try to derive it from other standard env variables that our users already use, possibly for APM?

For example [we have](https://docs.datadoghq.com/tracing/trace_collection/library_config/cpp/):
- `DD_AGENT_HOST`: example `localhost`.
- `DD_TRACE_AGENT_PORT`: default to `8126`.
- If present, both of them are overruled by `DD_TRACE_AGENT_URL`, which also requires schema and port, example `http://localhost:8126`

PR #113, opened 2024-09-18
chore(documentation): improvement of the documentation and the build process

dubloom:
Just by following the readme, it was almost impossible to either build something on Mac or even run the tests. 

From my point of view, when arriving on a repository should be able to build/test etc... the thing without modifying anything not explicitly mentioned in the documentation

Therefore, this pull request aims to enrich the readme so that the various actions can be carried out by following only the steps within it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,17 @@ NGINX_SRC_DIR ?= $(PWD)/nginx
 ARCH ?= $(shell arch)
 COVERAGE ?= OFF
 DOCKER_REPOS ?= public.ecr.aws/b1o7r7e0/nginx_musl_toolchain
+MACOS ?= false
+ifeq ($(MACOS), true)
+	PCRE2_PATH ?= /opt/homebrew/Cellar/pcre2/10.44
+	MAC_OPTIONS = -DCMAKE_C_FLAGS=-I$(PCRE2_PATH)/include/ -DCMAKE_CXX_FLAGS=-I$(PCRE2_PATH)/include/ -DCMAKE_LDFLAGS=$(PCRE2_PATH)/lib -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang
+endif
 
 SHELL := /bin/bash
 
 .PHONY: build
-build: build-deps sources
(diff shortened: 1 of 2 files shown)


dubloom:
I also feel that scripts should not be quoted alone to launch tests, as they require upstream steps before they can be launched. I haven't modified it since it may be useful but in my opinion it could confuse users.

Makefile
```diff
@@ -9,13 +9,17 @@ NGINX_SRC_DIR ?= $(PWD)/nginx
 ARCH ?= $(shell arch)
 COVERAGE ?= OFF
 DOCKER_REPOS ?= public.ecr.aws/b1o7r7e0/nginx_musl_toolchain
+MACOS ?= false
+ifeq ($(MACOS), true)
+	PCRE2_PATH ?= /opt/homebrew/Cellar/pcre2/10.44
+	MAC_OPTIONS = -DCMAKE_C_FLAGS=-I$(PCRE2_PATH)/include/ -DCMAKE_CXX_FLAGS=-I$(PCRE2_PATH)/include/ -DCMAKE_LDFLAGS=$(PCRE2_PATH)/lib -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang
```
dmehala, line 15: I understand your contribution. However you can not suppose every macos user will have PCRE2 installed at this very specific location.

README.md
```diff
@@ -88,17 +88,50 @@ Requirements:
...
+```shell
+MACOS=true NGINX_VERSION=1.25.2 make build
+```
 
+You will need pcre2 to be able to build the module correctly.
+You can install it with:
 ```shell
-WAF=ON NGINX_VERSION=1.25.2 make build
+brew install pcre2
 ```
```
dmehala, line 126: PCRE2 is not a requirement on our module. However, the library is required when NGINX is built with some modules, e.g. the HTTP rewrite module. For example building NGINX with `--without-http_rewrite_module` remove the requirement for PCRE.

Your addition is misleading. In the other hand, I agree the current setup with two different build system is not ideal. I would prefer a troubleshooting section for common errors explaining the status and how to solve the issue.

dubloom:
Converted to draft because i will need to furthermore improve the doc after openresty support

PR #114, opened 2024-09-19
build(test): restrict gdb-15 install to amd64 architecture

dmehala:
# Description

PR #71 updates gdb to a newer version that supports reading DWARF debugging format. However, it incorrectly attempts to install the x86_64 version regardless of the system's architecture, causing errors on non-x86_64 platforms.

This fix temporarily ensures that the updated gdb is only installed on x86_64 (amd64) systems.

## Note to reviewers
This is blocking the new release.


test/services/nginx/install_tools.sh
```diff
@@ -24,11 +38,14 @@ if command -v apt-get >/dev/null 2>&1; then
     fi
 
     if grep ^11\. /etc/debian_version; then
-        # the version of gdb is too old and can't handle dwarf data generated by clang 16
-        apt-get install -y wget
-        wget -O /tmp/gdb.deb https://sqreen-ci-java.s3.amazonaws.com/deb-ad-hoc/bullseye/gdb_15.1-1_amd64.deb
-        dpkg -i /tmp/gdb.deb
-        rm /tmp/gdb.deb
+        ARCH="$(get_architecture)"
+        if [ "$ARCH" == "amd64" ]; then
```
dmehala, line 42:
```suggestion
        if [ "$ARCH" = "amd64" ]; then
```

dubloom:
It seems that we don't want to do anything on non amd64 system, do you know why ?

dmehala:
@dubloom gdb is installed for all architecture, but gdb-15 will not be installed for ubuntu on arm64 architecture for the following reasons:
1. It is currently blocking the release, so this serves as more of a workaround.
2. gdb is not mandatory for testing purpose, so this workaround should be effective.
3. The resources to build and package gdb15 for ubuntu arm64 are unfortunately unavailable at this time.
4. It's likely that the versions requiring a gdb upgrade will no longer be supported after this release.

pablomartinezbernardo (approved):

PR #127, opened 2024-10-03
Installer telemetry

pablomartinezbernardo:
Add installer telemetry following spec

image
image
image
image

### DIFF
--- a/installer/configurator/configurator.go
+++ b/installer/configurator/configurator.go
@@ -4,11 +4,14 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	log "github.com/sirupsen/logrus"
 )
 
-var InstallerVersion = "0.1.1"
+var InstallerVersion = "0.1.2"
+var IntegrationVersion = "unset"
+var IntegrationName = "unset"
 
 func validateInput(appID, site, clientToken, arch string, sessionSampleRate, sessionReplaySampleRate int) error {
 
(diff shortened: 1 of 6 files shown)


installer/configurator/telemetry_sender.go
dmehala, line 1: [naive question] Would it possible to reuse the Go tracer code?

pablomartinezbernardo: There's some drawbacks of doing that
* Forces installer to send the `app-started` which doesn't fit
* The version is always set to the version of the tracer
* We would have to be very careful to set appropriate values to skip certain parts that are tracer-specific. For example, it is trying to do dependency collection, which can only be disabled by setting DD_TELEMETRY_DEPENDENCY_COLLECTION_ENABLED to false, otherwise this client will attempt dependency collection
* Doesn't support telemetry logs

And probably some other difficulties that would arise once we started actually using it. All in all, I believe trying to use that client would make this code more confusing than just reimplementing the few things we are actually using.

installer/configurator/configurator.go
```diff
@@ -35,15 +38,51 @@ func validateInput(appID, site, clientToken, arch string, sessionSampleRate, ses
...
+					"integration_version": IntegrationVersion}),
+			}); err != nil {
+				log.Error("Error sending error log: ", err)
+			}
+
+			if err := sender.SendInstallationCount("error", errorType.String()); err != nil {
+				log.Error("Error sending error count: ", err)
+			}
+
+			end := time.Now().UnixNano() / int64(time.Millisecond)
```
dmehala, line 70: [question] Can `start` and `end` be in Nanoseconds and convert in millisecond while computing the delta?

installer/configurator/configurator.go
```diff
@@ -69,35 +110,51 @@ func main() {
...
 		log.Info("Datadog NGINX module has been successfully installed and configured. Please reload NGINX or restart the service for the changes to take effect")
 	}
+
+	if !*dryRun {
+		log.Debug("Sending installation telemetry")
+		if err := sender.SendInstallationCount("success", ""); err != nil {
+			log.Error("Error sending installation success count: ", err)
+		}
+
+		end := time.Now().UnixNano() / int64(time.Millisecond)
```
dmehala, line 154: [nit] Probably don't make a huge difference, however, I would argue to set `end` right after `if !*dryRun` for accuracy

installer/configurator/telemetry_sender.go
```diff
@@ -0,0 +1,278 @@
...
+func initApplication(serviceName, env string) (Application, error) {
+	if serviceName == "" || env == "" {
+		return Application{}, NewInstallerError(TelemetryError, fmt.Errorf("serviceName and env are required"))
+	}
+
+	return Application{
+		ServiceName:     serviceName,
+		Env:             env,
+		ServiceVersion:  InstallerVersion,
+		TracerVersion:   InstallerVersion,
```
dmehala, line 119: [recommendation] If we just use the installer version, I am afraid it will be confusing for one to know if it's actually a tracer version or something else. How about we prefix this with "nginx-installer-"?

dmehala (commented):
I added few questions. Would it be possible to add tests for telemetry?

installer/configurator/telemetry_sender.go
pablomartinezbernardo: It's already part of the service name, so it wouldn't be adding any extra information

dmehala (approved):
LGTM! :shipit:

PR #128, opened 2024-10-16
build: support ingress-nginx

dmehala:
# Description

Generate compatible module that can be loaded by ingress-nginx controller.

Changes:
  - Refactor `bin/release.py` to handle releasing the ingress-nginx init container.
  - Add `bin/ingress-nginx.py` script to build a module version compatible with the ingress-nginx controller.
  - Add CI jobs for building and testing ingress-nginx modules.
  - Add a new Makefile target to build the ingress-nginx module.
  - Include documentation and examples for using ingress-nginx.
  - Skip features incompatible with the ingress-nginx controller.
  - Update tests to remove the `root` directive, which is not supported by ingress-nginx.
  - Modify `orchestrator.py` to restart the nginx service using the default config.
  - Enforce the creation of a PID file for the nginx service.


dmehala:
Work as expected, however, we are tracing healthcheck:
Screenshot 2024-10-16 at 15 46 39

Solution: drop-in support.

example/ingress-nginx/helm/values.yaml
```diff
@@ -0,0 +1,20 @@
...
+  opentelemetry:
+    enabled: false
+  extraModules:
+    - name: nginx-datadog
+      image:
+        registry: docker.io
+        image: datadog/ingress-nginx-injection
+        # The tag should match the version of the ingress-nginx controller
+        # For example, this will inject the Datadog module for ingress v1.10.0
+        # Check <> for the list of all versions supported.
```
pablomartinezbernardo, line 13:
```suggestion
        # Check <https://hub.docker.com/repository/docker/datadog/ingress-nginx-injection/tags> 
        # for the list of all versions supported.
```

bin/release.py
```diff
@@ -1,5 +1,18 @@
...
+Usage:
+======
+To release nginx-datadog modules:
+./release.py nginx-module --ci-token <CI_TOKEN> --version-tag v1.3.1 --workflow-id <CI_RELEASE_WORKFLOW_ID>
+
+To release ingress-nginx docker init containers:
+./release.py ingres-nginx --ci-token <CI_TOKEN> --version-tag v1.3.1 --workflow-id <CI_RELEASE_WORKFLOW_ID>
+
+To release the installer:
+./release.py ingres-nginx --ci-token <CI_TOKEN> --version-tag v0.3.3 --workflow-id <CI_RELEASE_WORKFLOW_ID>
```
pablomartinezbernardo, line 15:
```suggestion
./release.py installer --ci-token <CI_TOKEN> --version-tag v0.3.3 --workflow-id <CI_RELEASE_WORKFLOW_ID>
```

bin/release.py
```diff
@@ -208,83 +235,138 @@ def prepare_release_artifact(work_dir, build_job_number, version, arch, waf):
...
-                sys.exit(1)
-
         for job in jobs:
-            result = handle_job(job, Path(work_dir), options.installer)
-            if result != "done":
-                sys.exit(1)
+            # See the response schema for a list of statuses:
+            # https://circleci.com/docs/api/v2/index.html#operation/listWorkflowJobs
+            if job["name"].startswith("build ") and not job["name"].startswith(
+                    "build installer "):
```
pablomartinezbernardo, line 359:
```suggestion
            if job["name"].startswith("build ") and not job["name"].startswith(
                    "build installer ") and not job["name"].startsWith("build ingress-nginx"):
```

pablomartinezbernardo (approved):
🚀 !

.circleci/continue_config.yml
```diff
@@ -364,6 +451,42 @@ workflows:
...
+                        version:
+                            - v1.11.3
+                            - v1.11.2
+                            - v1.11.1
+                            - v1.11.0
+                            - v1.10.4
+                            - v1.10.3
+                            - v1.10.2
+                            - v1.10.1
+                            - v1.10.0
```
robmp, line 468: Would it be possible to build for v1.9.6 as well? This would make the transition easier, because it's the last version still supporting opentracing.

dmehala: Hey @robmp 
I don't quite understand the rationale behind building for pre v1.10. How would this help with the transition?

.circleci/continue_config.yml
```diff
@@ -97,6 +97,46 @@ jobs:
...
+        steps:
+            - checkout
+            - run: git submodule sync && git submodule update --init --recursive
+            - run:
+                command: |
+                    if [ "<< parameters.arch >>" == "amd64" ]; then
+                      echo 'export NGINX_ARCH=x86_64' >> $BASH_ENV;
+                    else
+                      echo 'export NGINX_ARCH=aarch64' >> $BASH_ENV;
+                    fi
```
dmehala, line 122:
```suggestion
```

bin/ingress_nginx.py
```diff
@@ -0,0 +1,205 @@
...
+        "v1.10.3": f"{ingress_rootdir}/images/nginx-1.25/rootfs/patches",
+        "v1.10.2": f"{ingress_rootdir}/images/nginx-1.25/rootfs/patches",
+        "v1.10.1": f"{ingress_rootdir}/images/nginx-1.25/rootfs/patches",
+        "v1.10.0": f"{ingress_rootdir}/images/nginx-1.25/rootfs/patches",
+    }
+    return mapping.get(version, "")
+
+
+def patch_nginx(src_dir: str, patch_dir: str) -> None:
+    for patch_file in os.listdir(patch_dir):
```
dmehala, line 86:
```suggestion
def patch_nginx(src_dir: str, patch_dir: str) -> None:
    if DRY_RUN:
       return
    for patch_file in os.listdir(patch_dir):
```

.circleci/continue_config.yml
robmp: We currently use opentracing with `datadog-collector-host` in our nginx ingress controllers, and we have many independently deployed ingresses which have opentracing enabled via the `enable-opentracing` annotation, and also have some additional opentracing configuration in `configuration-snippet` annotations.

Nginx ingress controller v1.10.0 drops support for opentracing, so in order to upgrade and switch to the datadog module, we would have to first remove all opentracing configuration from our ingresses, then upgrade the ingress controllers, then re-add the tracing configuration using the datadog module. This means we would have gaps in our tracing data during the transition, which is not ideal.

If the datadog module supported v1.9.6, we could switch the ingresses from opentracing to datadog one by one without any gaps in tracing data. The process would also be easier to test and roll back in case of any issues.

dmehala: Make sense, I added `v1.9.6` support in d9b8f77c06ed18b0867caca7f1d8b9a7dbff9cd0.

robmp: Thank you!

PR #137, opened 2024-11-04
feat: Support WAF address for request body

cataphract:
# Description
It intercepts the request body, parses it if it's json/multipart/urlencoded and sends it to the WAF, perhaps blocking afterwards.

It implements it by installing a request body filter that holds data (consumes and copies) until it either reaches a limit on the data we're willing to process or there's an EOF. When this happens, it calls the WAF in a separate thread, replaces the read/write handlers so there's no further async processing and does not consume the last set of buffers so the loop calling the filter exits, effectively suspending the request until waf completion. By that point, we either block, or, if we don't, we post a read event so the processing continues, and when the filter is called release downstream all the data held plus the new buffers gotten. Afterwards, we just pass through the data.

See [test_sec_addresses.py](https://github.com/DataDog/nginx-datadog/pull/137/files#diff-8e1be82682ad084cd3ea1231beb5fe8c460c87436de1e7bf2e0331f079c57797) (for instance) for sample test requests.

Related Jira: [APPSEC-53575]

[APPSEC-53575]: https://datadoghq.atlassian.net/browse/APPSEC-53575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dmehala:
Hey @cataphract 
Can you share more details on what this PR do and how to test it?

cataphract:
> Hey @cataphract Can you share more details on what this PR do and how to test it?

It intercepts the request body, parses it if it's json/multipart/urlencoded and sends it to the waf, perhaps blocking afterwards.

See [test_sec_addresses.py](https://github.com/DataDog/nginx-datadog/pull/137/files#diff-8e1be82682ad084cd3ea1231beb5fe8c460c87436de1e7bf2e0331f079c57797) (for instance) for sample test requests.

src/security/body_parse/generator.h
```diff
@@ -0,0 +1,71 @@
...
+      return;
+    }
+
+    if (h_.done()) {
+      std::terminate();
+    }
+
+    h_();
+    if (h_.promise().exception_)
+      std::rethrow_exception(h_.promise().exception_);
```
Anilm3, line 64:
```suggestion
    if (h_.promise().exception_) {
      std::rethrow_exception(h_.promise().exception_);
    }
```

I thought you had a clang-tidy check for this.

src/security/body_parse/generator.h
```diff
@@ -0,0 +1,71 @@
...
+  bool fetch_next_ = true;
+
+  void fill() {
+    if (!fetch_next_) {
+      return;
+    }
+
+    if (h_.done()) {
+      std::terminate();
+    }
```
Anilm3, line 60: Not sure about this, I'm struggling to see why the need to call `std::terminate` here...

Anilm3 (commented):
I'm still wrapping my head around some parsers, I'll continue soon...

src/security/ddwaf_memres.h
```diff
@@ -96,4 +96,47 @@ class DdwafMemres {
...
 
+/*
+ * A pool of ddwaf object vectors of sizes in powers in two. Used when the
+ * sizes of arrays or maps are not known in advance.
+ */
+template <typename DdwafObjType>
+requires std::is_base_of_v<ddwaf_object, DdwafObjType>
+class DdwafObjArrPool {
+ public:
+  DdwafObjArrPool(DdwafMemres &memres) : memres_{memres} {}
```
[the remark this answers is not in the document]
cataphract: the memory returned by `DdwafObjArrPool` is owned `DdwafMemres` (which is owned by the `Context`); since the pool is used to build data submitted to the waf, it needs to be live until the waf context is destroyed.

Note that `DdwafObjArrPool` is destroyed by the time `parse_json` / `parse_multipart` return; it's merely a helper to use the `DdwafMemres` more efficiently.

src/security/body_parse/body_parsing.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+    // count key occurrences
+    std::unordered_map<std::string_view, std::size_t> bag;
+    for (; !it.ended(); ++it) {
+      std::string_view cur_key = it.cur_key();
+      bag[cur_key]++;
+    }
+
+    // allocate all ddwaf_obj, set keys
+    std::unordered_map<std::string_view, ddwaf_obj *> key_index;
```
Anilm3, line 122: You could use the same map for both key counting and keeping track of duplicate keys, just replace the value with `std::pair<std::size_t, ddwaf_obj *>` or, better yet, a struct.

src/security/body_parse/body_parsing.cpp
```diff
@@ -0,0 +1,158 @@
...
+    for (it.reset(); !it.ended(); ++it) {
+      auto [cur_key, cur_value] = *it;
+      ddwaf_obj &cur = *key_index.at(cur_key);
+      if (cur.is_string()) {
+        cur.make_string(cur_value);
+      } else {
+        ddwaf_arr_obj &cur_arr = static_cast<ddwaf_arr_obj &>(cur);
+        cur_arr.at_unchecked(cur_arr.nbEntries++).make_string(cur_value);
+      }
+    }
```
Anilm3, line 148: Could be done in a single pass.

src/security/body_parse/header.cpp
```diff
@@ -0,0 +1,520 @@
...
+      "abcdefghijklmnopqrstuvwxyz"
+      "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+      "0123456789!#$%&'*+-.^_`|~";
+  auto end = sv.find_first_not_of(tchar);
+  if (end == 0) {
+    return std::nullopt;
+  }
+  if (end == std::string_view::npos) {
+    end = sv.size();
+  }
```
Anilm3, line 82: This is not needed, when the `count` passed to `substr` is `npos`, the substring will be from `start` until the end of the string.

src/security/body_parse/header.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+/*
+ * https://httpwg.org/specs/rfc9110.html#quoted.strings
+ * quoted-string  = DQUOTE *( qdtext / quoted-pair ) DQUOTE
+ * qdtext         = HTAB / SP / %x21 / %x23-5B / %x5D-7E / obs-text
+ * obs-text       = %x80-FF
+ * quoted-pair    = "\" ( HTAB / SP / VCHAR / obs-text )
+ */
+std::optional<std::string> consume_9110_quoted_string(std::string_view &sv) {
+  if (sv.front() != '"') {
```
Anilm3, line 96: Best to use `starts_with` as it'll do an empty check as well.

src/security/body_parse/header.cpp
```diff
@@ -0,0 +1,520 @@
...
+    // qdtext
+    if (ch == '\t' || ch == ' ' || (ch >= 0x21 && ch != '\\' && ch != 0x7F)) {
+      result.push_back(ch);
+      continue;
+    }
+
+    // quoted-pair
+    if (ch == '\\') {
+      if (sv.empty()) {
+        return std::nullopt;
```
Anilm3, line 118: More readable to just break, here and at the end of the loop.

src/security/body_parse/header.cpp
```diff
@@ -0,0 +1,520 @@
...
+    if (sv.empty()) {
+      return ct;
+    }
+
+    std::optional<std::string_view> maybe_param_name = consume_wg_token(sv);
+    if (!maybe_param_name) {
+      continue;
+    }
+
+    if (sv.empty() || sv.front() != '=') {
```
Anilm3, line 298: Could check if `sv.size() <= 1` and elide the next check after `remove_prefix`.

src/security/body_parse/body_json.cpp
```diff
@@ -0,0 +1,303 @@
...
+using dnsec::ddwaf_obj;
+
+/* Adapt a nginx chain to a rapidjson input.
+ * see https://rapidjson.org/classrapidjson_1_1_stream.html */
+class RapidNgxChainInputStream {
+ public:
+  using Ch = char;
+
+  RapidNgxChainInputStream(const ngx_chain_t *chain) : current_{chain} {
+    if (current_) {
```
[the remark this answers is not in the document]
cataphract: yeah I'm making it thow

src/security/body_parse/generator.h
cataphract: that's clang-format I think

src/security/body_parse/generator.h
cataphract: if we get here, we asked for another value after the `co_return`, which shouldn't happen (there's no value to give)

src/security/body_parse/body_parsing.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+bool parse_body(ddwaf_obj &slot, ngx_http_request_t &req,
+                const ngx_chain_t &chain, std::size_t size,
+                DdwafMemres &memres) {
+  if (is_json(req)) {
+    // use rapidjson to parse:
+    bool success = parse_json(slot, req, chain, memres);
+    if (success) {
+      return true;
+    }
```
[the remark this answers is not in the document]
cataphract: That's the idea yes. Maybe the downstream server doesn't even check the content-type. It's a debatable choice for sure.

src/security/body_parse/body_parsing.cpp
cataphract: I don't see how. On the first pass we're counting total occurrences of a key, which we need to decide the size of the array or if we just need to allocate for a string ddwaf_object. To do it on one single pass, we'd need to do reallocations, which certainly worse in the worst case scenario, though maybe not in the most typical ones.

src/security/body_parse/header.cpp
cataphract: there may be other advantages to avoid early returns, but I don't think readability is one them. This way you know right there what is happening, rather than scrolling down to follow the flow and find out

src/security/body_parse/body_parsing.cpp
Anilm3: I meant specifically those two loops, the counting loop happens before these two.

src/security/body_parse/body_parsing.cpp
Anilm3: It might be more confusing but it avoids the generation of the same hash table twice.

src/security/body_parse/body_parsing.cpp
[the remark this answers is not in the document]
Anilm3: The point I was trying to convey (poorly) is that the raw body shouldn't be provided in the `server.request.body` address.

src/security/body_parse/header.cpp
Anilm3: It's effectively the same argument as goto cleanup.

src/security/body_parse/body_parsing.cpp
cataphract: ok i put a union in place

src/security/body_parse/body_parsing.cpp
cataphract: they're iterating different things. The first loop will do one iteration per each key, whereas the second will do one per key occurrence. I suppose it is possible, to merge the first loop into the second if you had a flag to tell you whether it's the first time the value was seen, the result is more memory and code that's imo less clear.

src/security/body_parse/body_parsing.cpp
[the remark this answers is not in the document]
cataphract: ok I see. So I should provide it only if the content-type is text/plain?

src/security/body_parse/header.cpp
cataphract: Yes, but `end` is also used on the `remove_prefix` call, which doesn't accept npos

src/security/body_parse/body_parsing.cpp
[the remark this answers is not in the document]
Anilm3: Yeah, I think that might be best.

src/security/body_parse/body_parsing.cpp
Anilm3:
```cpp

    for (it.reset(); !it.ended(); ++it) {
      auto [cur_key, cur_value] = *it;
      auto key_index_it = key_index.find(cur_key);
      if (key_index_it == key_index.end()) {
         ddwaf_obj &cur = slot_map.at_unchecked(i++);
         auto [new_it, res] = key_index.emplace(key, &cur);
         cur.set_key(key);

         auto count = bag[cur_key]; // Assume it's present
         if (count == 1) {
           cur.make_string(""sv);  // to be filled later
         } else {
           cur.make_array(count, memres);
           cur.nbEntries = 0;  // fixed later
         }
         key_index_it = new_it;
      }
      
      ddwaf_obj &cur = key_index_it->second;
      if (cur.is_string()) {
        cur.make_string(cur_value);
      } else {
        ddwaf_arr_obj &cur_arr = static_cast<ddwaf_arr_obj &>(cur);
        cur_arr.at_unchecked(cur_arr.nbEntries++).make_string(cur_value);
      }
    }
```

I haven't checked if it works or not, I just roughly copy-pasted to give you an idea:
- You could even reduce lookups by accessing `key_index` with `operator[]` and checking for `nullptr`,  which would allow you to avoid the extra emplace.
- The initialisation could also set the values instead of creating those temporary empty versions.

 Admittedly it's not much easier to read, but it might be simpler overall.

cataphract: ok but in this case you're assuming I didn't heed your previous recommendation if merging the two hash maps

Anilm3: Yes, I didn't account for that.

src/security/body_parse/header.cpp
Anilm3: Ah yes, I missed that entirely.

src/security/body_parse/generator.h
Anilm3: Would it not be better to throw? It shouldn't happen anyway but if there's a bug somewhere (as it has happened in the past) you can at least recover.

Anilm3 (approved):
lgtm, great work!

dmehala (commented):
I do have a few questions regarding buffer management, especially, taking ownership of buffer seems problematic but may lack some knowledge here.

There is a lot of customer parser involved, I would strongly suggest to write fuzzer for those - especially the multipart and urlencoder parser - before merging the PR.

Additionnally, I would like your opinion on a better way to manager the state machine. Right now, the state is bleeding everywhere and is challenging to track. How about one big function to manage the state machine and dispatch? Something like:

```cpp
enum ngx_handler {
    log_request,
    ...
    header_filter
};

template <typename T, typename... Args>
void Context:on_handler(ngx_handler handler, Args&&... args) {
    switch(state_) {
        case stage::BEFORE_RUN_WAF_END:
            if (handler == ngx_handler::header_filter) {
                auto rc = on_header_filter(std::forward<Args>(args)...);
                if (rc == OK) transition_to_stage(stage::XYZ);
            }
        ...
        default:
            ngx_log_error("wrong state for handler XYZ")
    }
}

// in datadog_context.cpp
ngx_int_t on_log_request(ngx_http_request_t* request) {
    sec_ctx_->on_handler(ngx_handler::log_request, request);
}
```

CMakeLists.txt
```diff
@@ -227,4 +250,7 @@ if(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo" AND CMAKE_SYSTEM_NAME STREQUAL "Li
   split_debug_info(ngx_http_datadog_module)
 endif()
 
+add_library(ngx_http_datadog_static_lib STATIC EXCLUDE_FROM_ALL)
+target_link_libraries(ngx_http_datadog_static_lib PUBLIC ngx_http_datadog_objs)
+
```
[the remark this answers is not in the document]
cataphract: Yes, it's used for linking the unit_tests against:

```
test/unit/CMakeLists.txt:14:    target_link_libraries(unit_tests PRIVATE ngx_http_datadog_static_lib Catch2::Catch2WithMain libddwaf_objects)
```

src/security/context.cpp
```diff
@@ -643,16 +694,197 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+  ngx_log_debug3(
+      NGX_LOG_DEBUG_HTTP, request.connection->log, 0,
+      "request body filter %s in chain. Current data: %lu, Stage: %d",
+      in ? "with" : "without", filter_ctx_.out_total, st);
+
+  if (st == stage::AFTER_BEGIN_WAF) {
+    ngx_log_debug1(NGX_LOG_DEBUG_HTTP, request.connection->log, 0,
+                   "first filter call, req refcount=%d", request.main->count);
+    // buffer the request body during reading
+    // https://github.com/nginx/nginx/commit/67d160bf25e02ba6679bb6c3b9cbdfeb29b759de
```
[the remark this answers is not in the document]
cataphract: The lowest version we build for 1.22.0. Are we supposed to support older versions? Because I can either disable this feature for nginx < 1.21.4 or just note in the README

src/security/context.cpp
```diff
@@ -643,16 +694,197 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+    for (auto *cl = in; cl; cl = cl->next) {
+      new_size += cl->buf->last - cl->buf->pos;
+      is_last = is_last || cl->buf->last_buf;
+    }
+
+    bool run_waf = new_size >= kMaxFilterData || is_last;
+    if (run_waf) {
+      // do not consume the buffer so that this filter is not called again
+      if (buffer_chain(*request.pool, in, false) != NGX_OK) {
+        return NGX_HTTP_INTERNAL_SERVER_ERROR;
```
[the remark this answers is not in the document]
cataphract: this can only happen on OOM conditions, when `malloc` returns null. At that point, it's not likely there's anything working.

src/security/context.cpp
```diff
@@ -643,16 +694,197 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+      // pass saved buffers downstream
+      auto rc = ngx_http_next_request_body_filter(&request, filter_ctx_.out);
+
+      for (auto *cl = filter_ctx_.out; cl;) {
+        auto *ln = cl;
+        cl = cl->next;
+        ngx_free_chain(request.pool, ln);
+      }
+      filter_ctx_.out = nullptr;
+      filter_ctx_.out_last = &filter_ctx_.out;
```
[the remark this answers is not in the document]
cataphract: `FilterCtx` has no destructor, so the data inside it wouldn't be freed by the destruction of the context; however, it the chain links are allocated as part of the request pool and would be freed then anyway. In general it's best to return data used to the request pool so it can be reused. I do notice though, that I don't free the temp buffers allocated in `buffer_chain` when `consume == true`; although not critical, that could be improved here.

src/security/context.cpp
```diff
@@ -643,16 +694,197 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+      }
+
+      new_buf->last = ngx_copy(new_buf->pos, buf->pos, size);
+      buf->pos = buf->last;
+      new_buf->last_buf = buf->last_buf;
+      new_buf->tag = buf->tag;
+
+      new_ch->buf = new_buf;
+    } else {
+      new_ch->buf = buf;
```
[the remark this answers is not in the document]
cataphract: The only situation where we don't consume the buffer (consume: false) is on the first call after the WAF is completed, and we immediately pass a chain with this unconsumed buffer downstream, followed by  releasing the whole chain. So in effect, as far as this buffer is concerned, it's as if we just directly passed downstream the chain gotten by the filter, unaltered (actually we join the in chain with the copy of all the other buffers we got before calling the WAF, but that doesn't matter).

Now, there is one extra situation where we don't consume the buffer: if we're called while the WAF is suspended: but if this happens, it's a bug. As to your point about buffer reuse: by definition, a buffer that's not consumed (where pos != last) is a busy buffer, so it cannot be moved to the freelist.

src/security/context.cpp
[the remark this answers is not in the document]
dmehala: > Are we supposed to support older versions?
No, we officially support only the last 2 years of NGINX. You'r good, please juste add a note in the README.

src/security/body_parse/body_parsing.cpp
```diff
@@ -0,0 +1,167 @@
...
+      return false;
+    }
+
+    return parse_multipart(slot, req, *ct, chain, memres);
+  }
+
+  bool ct_plain = is_text_plain(req);
+  bool ct_urlencoded = !ct_plain && is_urlencoded(req);
+
+  if (!ct_plain && !ct_urlencoded) {
```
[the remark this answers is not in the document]
cataphract: the content-type cannot be plain and urlencoded at the same time; logically `bool ct_urlencoded = !ct_plain && is_urlencoded(req);` is equivalent to the simpler `bool ct_urlencoded = is_urlencoded(req);`. However, the version with `!ct_plain` avoids doing the comparison if we already now it's text/plain. It's basically a (marginal) optimization.

src/security/context.cpp
```diff
@@ -643,16 +694,197 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+    size_t new_size = filter_ctx_.out_total;
+    bool is_last = filter_ctx_.found_last;
+    for (auto *cl = in; cl; cl = cl->next) {
+      new_size += cl->buf->last - cl->buf->pos;
+      is_last = is_last || cl->buf->last_buf;
+    }
+
+    bool run_waf = new_size >= kMaxFilterData || is_last;
+    if (run_waf) {
+      // do not consume the buffer so that this filter is not called again
```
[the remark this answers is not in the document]
cataphract: From what I can see this can only potentially be a problem if we get data while suspended (should not happen, and if it does already causes synchronisation issues and possible corruption). Otherwise the data is either copied and consumed or passed downstream immediately (where it will likely get to `ngx_http_request_body_save_filter` and be consumed there).

I guess I could add a condition to make it consume the buffer if the `recycled` flag is true, but if I need to get the data-on-suspended-state anomalous situation to work correctly I also need to add a separate chain to avoid synchronization issues.

src/security/context.cpp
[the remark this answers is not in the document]
dmehala: > FilterCtx has no destructor, so the data inside it wouldn't be freed by the destruction of the context

That's my point, this logic is better located in the `FilterCtx` destructor and the memory will be free on the non critical path. Or best, let the pool manage that.

src/security/context.cpp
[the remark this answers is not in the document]
dmehala: The **current** usage might be correct, but the implementation is not. While everything is fresh in your mind now and you understand all the interactions, consider the future. Think about the future you or a new contributor, who might need to modify this part of the module. Let's ensure it's as clear and correct as possible for them.

src/security/body_parse/generator.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+// adapted from: https://en.cppreference.com/w/cpp/language/coroutines
```
dmehala, line 3:
```suggestion
// adapted from: <https://en.cppreference.com/w/cpp/language/coroutines>
```

src/security/body_parse/generator.h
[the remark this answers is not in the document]
dmehala: Still doesn't hurt to add a few comment lines explaining what's the purpose of a Generator.

src/security/body_parse/body_parsing.cpp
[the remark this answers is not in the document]
dmehala: > the content-type cannot be plain and urlencoded at the same time;

That being said, may I ask, why not just `bool ct_urlencoded = is_urlencoded(req);` then? That's details but I am not sure there is any optimization here.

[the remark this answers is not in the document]
dmehala: Alright, just for the record, because I was looking for the optimization somewhere else, the optimization is not calling `is_urlencoded(req)` if it's plain text.

src/security/context.cpp
```diff
@@ -688,6 +920,53 @@ ngx_int_t Context::do_output_body_filter(ngx_http_request_t &request,
...
+  DDWAF_RET_CODE const code = ddwaf_run(ctx_.resource, &input, nullptr, &result,
+                                        Library::waf_timeout());
+  if (code == DDWAF_MATCH) {
+    results_.emplace_back(result);
+  } else {
+    ddwaf_result_free(&result);
+  }
+
+  std::optional<BlockSpecification> block_spec;
+  ddwaf_map_obj actions_arr{result.actions};
```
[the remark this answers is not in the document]
cataphract: The code was correct because actually `ddwaf_result_free` resets `result.actions` to an empty map, but I'm rewriting the code so it's not surprising.

src/security/context.cpp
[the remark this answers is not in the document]
cataphract: Like I said in the last paragraph, buffer reuse is not a problem: if the buffer is busy it can't be reused (`ngx_chain_update_chains` moves buffers to free only if they're not busy). More relevant is your other point about recycled buffers in the other comment; those must be consumed. I'm not sure if those are ever produced by a request body filter in an nginx module, but it's indeed better be safe.

src/security/context.cpp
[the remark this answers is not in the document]
cataphract: A couple of points:

* I was mistaken about the buffers; those can't be returned to the pool, only possibly reused by keeping freelists (e.g. with `ngx_chain_update_chains`). We're not interested in reusing buffers though, because we have an initial phase where we only allocate buffers and then after we pass them downstream they're not needed anymore. That is, we don't have allocations after the deallocations. So the memory occupied by those buffers can only be reclaimed when the pool is destroyed on request end.
* The chain links CAN be returned to the pool though, as they have a fixed size and `ngx_pool_t` keeps a freelist for them. If we released the chain links only at the end of the request, we'd be increasing memory consumption unnecessarily.

dmehala (approved):
LGTM :shipit: I highly suggest fuzz-tests 👀

cataphract:
@dmehala I added a fuzzer for multipart. It didn't find anything interesting.

As to your point about the stages: you're right that it's getting tricky; I decided to document the current state of affairs and I actually found two problems. Your proposal though while with merit would require more thought (e.g. different handlers take different arguments, so calls inside the template would be tricky -- prob the handler type would need to be a non-type template parameter) and intrusive changes in the codebase.

PR #142, opened 2024-12-02
feat(rum): Add support for array settings

dmehala:
# Description

Far from being the ideal solution but I lack time 🤷‍♂️

### DIFF
--- a/src/rum/config.cpp
+++ b/src/rum/config.cpp
@@ -11,21 +11,22 @@
 namespace {
 
 char *set_config(ngx_conf_t *cf, ngx_command_t *command, void *conf) {
-  auto *rum_config =
-      static_cast<std::unordered_map<std::string, std::string> *>(conf);
+  auto &rum_config =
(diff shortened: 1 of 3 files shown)


src/rum/config.cpp
```diff
@@ -11,65 +11,72 @@
...
   rapidjson::Document doc;
   doc.SetObject();
 
   rapidjson::Document::AllocatorType &allocator = doc.GetAllocator();
   doc.AddMember("majorVersion", config_version, allocator);
 
   rapidjson::Value rum(rapidjson::kObjectType);
-  for (const auto &[key, value] : config) {
-    if (key == "majorVersion") {
-      continue;
```
pablomartinezbernardo, line 60: Curious to know: Is there a reason why we were skipping majorVersion before?

pablomartinezbernardo (approved):

src/rum/config.cpp
dmehala: one of the first iteration, `majorVersion` was part of the map but I quickly removed it to be a seperate argument. I think it is a left over.

PR #145, opened 2024-12-06
Add openresty compatibility for the nginx module

dubloom:
Added build step for build the model again OpenResty. 

We are now building using musl-libc and with no new Docker image. The base image is therefore nginx_musl_toolchain like the base build.
OpenResty patches the code of nginx. A openresty docker image exists but the patched code is deleted. Therefore we need to build openresty from scratch when we want to build the module which is quite long and will probably **significantly increase** the duration of the CI. I'm in favor of uploading image with openresty prebuilt without deleting the nginx code for the CI 

**Some paths differ between nginx and openresty in test cases, to solve that:**
We create two folders a base conf for nginx and one for openresty and when we know which one will be needed we copy it in the right folder

**If you want to test locally**
You need to create a openresty-version-info.
Then you can run:
````
WAF=ON ARCH=aarch64 RESTY_VERSION=1.27.1.1 make test-openresty
````

**The documentation will come with the Documentation refactoring PR**

### CODE REVIEW

**Comment by dubloom** [comment]
@dmehala I made a refactoring which i thought would be harmless before the creation of the PR. Should be fixed now

**Review by dmehala — COMMENTED** [review]
I confirm it is working, thank you @dubloom. I think this can be greatly simplified, dependencies (openssl and pcre2) don't need to build patched and built. Have a look [here](https://github.com/DataDog/nginx-datadog/compare/dubloom/openresty-build...dmehala/openresty-support). 

Please make sure to address the CI issues.

**Comment by dubloom** [comment]
@dmehala are you sure about the simplifications ? Each version of openresty use a specific version of openssl and pcre. We could lose the exact behavior of the exact version openresty by removing this part of the build

File: bin/openresty/build_common.sh
Comment (dmehala) [inline bin/openresty/build_common.sh:1]: I reiterate here just so you can track unnecessary changes. Preparing openresty can be greatly simplified and I doubt there's a need for one script per sets of versions.

File: lab/bin/run
Comment (dmehala) [inline lab/bin/run:1]: No need to update the `lab` folder. Outside of David I don't think anybody uses it.

File: openresty-version-info.example
Comment (dmehala) [inline openresty-version-info.example:1]: [nginx-version-info.example](https://github.com/DataDog/nginx-datadog/blob/master/nginx-version-info.example) is from a time when it was way more challenging to run the tests. I am actually planning to remove this file, in that regards, I think it is also not needed to replicate it for openresty.

File: test/bin/run [hunk 1]
```diff
@@ -4,7 +4,13 @@ cd "$(dirname "$0")/../"
...
 set -e
 
-. ../nginx-version-info
+if [ $(echo ${RESTY_TEST}) = "ON" ] ; then
+    . ../openresty-version-info
+    cp ../test/cases/auto_propagation/openresty_base_conf/* ../test/cases/auto_propagation/conf/
+else
+    . ../nginx-version-info
+    cp ../test/cases/auto_propagation/nginx_base_conf/* ../test/cases/auto_propagation/conf/
+fi
```
dmehala, line 13: [suggestion] I would suggest a solution that do not mess with the user filesystem.

test/bin/run
```diff
@@ -13,3 +19,4 @@ printf " test args : %s\n" "$@"
 
 docker compose build --parallel --progress=plain
 python3 -m unittest "$@"
+# docker-compose up --abort-on-container-exit
```
dmehala, line 22: [question] left over?

test/cases/auto_propagation/conf/uwsgi_auto.conf
```diff
@@ -12,7 +12,7 @@ http {
         listen       80;
 
         location / {
-            include /etc/nginx/uwsgi_params;
+            include /usr/local/openresty/nginx/conf/uwsgi_params;
```
dmehala, line 15: [suggestion] May I suggest another approach? Update the service [Dockerfile](https://github.com/DataDog/nginx-datadog/blob/master/test/services/nginx/Dockerfile) to create a symlink from `/usr/local/openresty/nginx/conf/uwsgi_params` to `include /etc/nginx/uwsgi_params`.

test/cases/orchestration.py
```diff
@@ -726,8 +726,6 @@ def reload_nginx(self, wait_for_workers_to_terminate=True):
                 "--",
                 "nginx",
                 "nginx",
-                "-g",
-                "pid /run/nginx.pid;",
```
dmehala, line 730: [question] May I ask why?

dmehala (commented):
Thank you for addressing my previous concerns. I left a few comments worth addressing. Please also make sure to:
- Add an openresty example in the [example folder](https://github.com/DataDog/nginx-datadog/tree/master/example).
- CI is green on all jobs.
-

test/cases/orchestration.py
dubloom: When we build a openresty container, there is no /run/nginx.pid file, therefore all the tests are failing. Furthemore, i found out that removing this lines does not change anything to nginx tests. Therefore i just removed them

test/bin/run
dubloom: yes

lab/bin/run
dubloom: done (by deleted it on my branch)

test/cases/orchestration.py
dmehala: This is mandatory for ingress-nginx which still passes for some reason on this branch. I noticed some errors in the test.

dubloom: Ok. I'll try to find another way

dubloom:
Notes: I deleted the nginx/openresty-info. Now the base_image is passed either from BASE_IMAGE make test or deduce from the NGINX/OPENRESTY_VERSION. A PR based on this branch and improving the documentation will be open soon

dubloom:
Note: the openresty-example will eventually disappear but as no version has been currently released so this is the only way to launch it

lab/bin/run
dmehala, line 1: This reintroduce [the recently deleted](https://github.com/DataDog/nginx-datadog/pull/146) `fab` folder

example/openresty/README.md
dmehala, line 1: [nitpick] Looks like this example is a copy paste from the other example. I would prefer simpler example.

bin/openresty/build_openresty.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
+
+export RESTY_J="-j8"
+export _RESTY_CONFIG_DEPS=""
+export RESTY_CONFIG_OPTIONS=""
+export RESTY_LUAJIT_OPTIONS=""
+export RESTY_PCRE_OPTIONS=""
```
dmehala, line 9: These variables are empty—what's the purpose?

bin/generate_jobs_yaml.rb
```diff
@@ -32,14 +32,29 @@
...
 TAB
 
+resty_version_table = <<-TAB
+openresty/openresty:1.27.1.1-alpine 1.27.1.1
+openresty/openresty:1.25.3.2-alpine 1.25.3.2
+openresty/openresty:1.25.3.1-alpine 1.25.3.1
+openresty/openresty:1.21.4.4-alpine 1.21.4.4
+openresty/openresty:1.21.4.3-alpine 1.21.4.3
+openresty/openresty:1.21.4.2-alpine 1.21.4.2
+openresty/openresty:1.21.4.1-alpine 1.21.4.1
```
dmehala, line 42: These are outdated versions of NGINX. Our policy supports only the four most recent versions. Could you remove these?

dubloom:
All the comments were taken into account except the two I answered.

bin/openresty/build_openresty.sh
dmehala: It's admirable to consider the future, but all too often the need is non-existent or is misaligned with twhat you'll actually require later. To keep things simple, I suggest removing this part. I'm perfectly okay if you choose to keep it, but I would kindly request that you document it at the very least (the why and how).

test/cases/openresty/test_openresty.py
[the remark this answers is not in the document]
dmehala: Oh I see, I missed that part. Well done.

bin/openresty/build_openresty.sh
dubloom: I removed it for now. If we need it later, we should be able to remember this conversation

PR #153, opened 2025-01-08
feat: support unified service tagging directive in server block

### LINKED ISSUE #147: Feature Request: Support setting service name per server
The `datadog_service_name` directive is the only way to override the default service name, 'nginx'. However, it is allowed only in the _http_ block, and not in the _server_ or _location_ blocks. So, spans emitted by all the servers in an nginx deployment with the nginx-datadog module end up belonging to the same service.

Being able to set different service names for different servers can be of great value. For instance, in cases where the nginx-datadog module is used with the ingress-nginx controller, this will allow setting service names on a per-ingress basis, which makes distinguishing between traffic flowing through different Ingresses easier and troubleshooting more effective.

ngx_http_opentracing_module allows setting service names on http, server, and location blocks with `opentracing_tag "service.name" "an_example_service";`. But, the following doesn't have the same effect with ngx_http_datadog_module. It merely adds it as span attribute.
```
datadog_tag "service.name"  "an_example_service";
```

dmehala:
Changes:
  - Allow `datadog_service_name` and `datadog_environment` in server block.
  - Add `datadog_version` directive.
  - Update documentation to reflect those changes.
  - Update integration tests.

Resolves #147


doc/API.md
```diff
@@ -51,18 +51,29 @@ The Datadog tracing module defines the following configuration directives.
 ### `datadog_service_name`
 - **syntax** `datadog_service_name <name>`
 - **default**: `nginx`
-- **context**: `http`
+- **context**: `http`, `server`
 
 Set the service name to associate with each span produced by this module.
+May overwrite `DD_SERV` environment variable.
```
zacharycmontoya, line 57: Should this be `DD_SERVICE`?

```suggestion
May overwrite `DD_SERVICE` environment variable.
```

test/cases/configuration/test_configuration.py
```diff
@@ -208,3 +239,50 @@ def run_precedence_test(self, conf_relative_path):
...
+            "service": "server1_block",
+            "environment": "server1",
+            "version": "1.5.0-server1",
+        }
+
+        mismatches = find_mismatches(pattern, config_srv2)
+        self.assertEqual(mismatches, [])
+
+        config_srv3 = json.loads(body3)
+        pattern = {
```
zacharycmontoya, line 281: Can we update this test case so that it only partially overrides the settings? Perhaps only update `datadog_service_name` and validate the output of that

src/request_tracing.cpp
```diff
@@ -191,6 +193,16 @@ RequestTracing::RequestTracing(ngx_http_request_t *request,
                  "starting Datadog request span for %p", request_);
 
   dd::SpanConfig config;
+  if (srv_conf_->service_name) {
```
zacharycmontoya, line 196: Do we have the capability to do trace stats calculations in the tracer? If we don't, then there are no potential issues. However, if we do then we will have to be careful about how we calculate if a span is "top-level" or not, now that we have the global tracer configuration set to service name `nginx` and the individual spans may receive a custom service name

zacharycmontoya (approved):
I left a couple of comments, but overall this looks good to me. Nice work 👍🏼

dubloom:
Reviewing now....

doc/API.md
```diff
@@ -51,18 +51,29 @@ The Datadog tracing module defines the following configuration directives.
 ### `datadog_service_name`
 - **syntax** `datadog_service_name <name>`
 - **default**: `nginx`
-- **context**: `http`
+- **context**: `http`, `server`
```
dubloom, line 54: is there a reason to not add the directive in location block ? I will add this question once but it is the same for the others directives

src/ngx_http_datadog_module.cpp
```diff
@@ -736,6 +762,27 @@ char *merge_script(ngx_conf_t *conf, NgxScript &previous, NgxScript &current,
...
+  auto child_srv_conf = static_cast<datadog_srv_conf_t *>(child);
+
+  if (!child_srv_conf->service_name) {
+    child_srv_conf->service_name = parent_srv_conf->service_name;
+  }
+  if (!child_srv_conf->service_env) {
+    child_srv_conf->service_env = parent_srv_conf->service_env;
+  }
+  if (!child_srv_conf->service_version) {
+    child_srv_conf->service_version = parent_srv_conf->service_version;
```
dubloom, line 780: Can't we use ngx_conf_merge_value here ?

test/cases/configuration/test_configuration.py
dubloom: 👍

dubloom (commented):
Not many things, just suggested the use of nginx_conf_merge if it applies here and I'm interested in knowing why you did not add the location context support ?

src/request_tracing.cpp
dubloom: ➕

dmehala: Care to explain?

test/cases/configuration/test_configuration.py
dmehala: This test assess the precedence of unified service tagging directives, rather than just the `datadog_service_name`. Partially overriding only `datadog_service_name` would defeat the purpose of the test. Am I misunderstanding something?

dubloom: My +1 was on the wrong comment (wanted to write here). I think what he meant was that it would be interesting that partial modification does not break the rest of the serv_conf.

doc/API.md
dmehala: addressed in b558dc79273d05d6ab4a26f008a79a1c7d78c812

doc/API.md
dmehala: addressed in b558dc79273d05d6ab4a26f008a79a1c7d78c812

src/ngx_http_datadog_module.cpp
dmehala: `ngx_conf_merge_value` only works for ngx types.

dmehala:
TODO: add `spanconfig` when the trace context is extracted.

src/datadog_directive.cpp
```diff
@@ -442,28 +442,95 @@ static char *set_configured_value(
...
+                  location.line);
+    return static_cast<char *>(NGX_CONF_ERROR);
+  }
+
+  loc_conf.service_env = {.location = location,
+                          .value = std::string(service_env)};
+  return NGX_OK;
+}
+
+char *set_datadog_version(ngx_conf_t *cf, ngx_command_t *command,
```
dubloom, line 505: Can't we refactor these 3 functions (maybe nginx does not allow that idk) ? It seems that the body is very alike in the three.

src/tracing_library.cpp
```diff
@@ -40,22 +37,13 @@ dd::Expected<dd::Tracer> TracingLibrary::make_tracer(
...
 
-  if (nginx_conf.service_name) {
-    config.service = nginx_conf.service_name->value;
-  } else {
-    config.service = "nginx";
-  }
-
-  if (nginx_conf.environment) {
-    config.environment = nginx_conf.environment->value;
-  }
```
dubloom, line 57: we stop setting up the environment because we are not sure anymore that it will be set, am i right ?

src/datadog_directive.cpp
dmehala: We can but feels more like a burden than providing additional value rn.

src/tracing_library.cpp
dmehala: 'cause no default environment and version.

PR #157, opened 2025-01-13
fix: apply sampling rules on resources correctly

dmehala:
Previously, sampling rules with resource name predicates were not applied. This is due to resource name set at the end of thre request processing pipeline, while sampling is computed earlier in the pipeline (during injection). As a result, rules with resource name criteria cannot match.

This fix ensures that span resource names are set before sampling rules are evaluated.

### DIFF
--- a/src/request_tracing.cpp
+++ b/src/request_tracing.cpp
@@ -195,6 +195,7 @@ RequestTracing::RequestTracing(ngx_http_request_t *request,
       to_system_timestamp(request->start_sec, request->start_msec);
   config.start = estimate_past_time_point(start_timestamp);
   config.name = get_request_operation_name(request_, core_loc_conf_, loc_conf_);
+  config.resource = get_request_resource_name(request_, loc_conf_);
 
   // By the end of this function, we will have a `request_span_`.
   //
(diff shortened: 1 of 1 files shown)


dmehala:
> Previously, sampling rules with resource name predicates were not applied. This is due to resource name set at the end of the request processing pipeline, while sampling is computed earlier in the pipeline (during injection). As a result, rules with resource name criteria cannot match.
> 
> This fix ensures that span resource names are set before sampling rules are evaluated.

From the PR message:
> Previously, sampling rules with resource name predicates were not applied. This is due to resource name set at the end of the request processing pipeline, while sampling is computed earlier in the pipeline (during injection). As a result, rules with resource name criteria cannot match.
> 
> This fix ensures that span resource names are set before sampling rules are evaluated.

[NGINX has a pipeline architecture for processing requests](https://nginx.org/en/docs/dev/development_guide.html#http_phases). When request comes in, span is created, injected and completed (setting `resource name` and other tags) at the very end of the pipeline. The fix actually set the `resource name` at the begining of the pipeline, which ensures the when sampling rules are evaluated the `resource name` is set.

dubloom:
Makes sense, thank you. What about that:
> However, in the code you deleted, it seems that some points were raised about a potential later modification. Do we know, if the new way of setting the resource_name takes that into account ?

PR #159, opened 2025-01-13
chore: update appsec rules to 1.13.3

dmehala:


Anilm3 (changes requested):
This looks like the ruleset version present in master rather than the actual 1.13.3 tag. I suggest not using this one as it has some stuff that may or may not be removed.

dmehala:
@Anilm3 Indeed, good catch!

Anilm3 (approved):
LGTM, thanks!

PR #170, opened 2025-02-19
build: add shim to `posix_spawn_file_actions_addchdir_np`

dmehala:


build_env/glibc_compat.c
```diff
@@ -87,137 +80,153 @@ int lstat(const char *restrict path, void *restrict buf)
...
+    posix_spawn_file_actions_t *restrict file_actions,
+    const char *restrict path) {
+  static int (*posix_spawn_file_actions_addchdir)(
+      posix_spawn_file_actions_t *restrict, const char *restrict);
+
+  if (!posix_spawn_file_actions_addchdir) {
+    (void)fprintf(stderr,
+                  "missing definition of `posix_spawn_file_actions_addchdir`");
+    abort();
+  }
```
cataphract, line 228: I don't see how this can work. `posix_spawn_file_actions_addchdir` is a pointer variable with internal linkage. It will not be touched by either the static or the dynamic linker and will be initialized to NULL. Since you never write it, the condition at line 224 will never be true and the program will always abort the first time is function is called.

Second, neither glibc nor musl appear to currently define `posix_spawn_file_actions_addchdir`. This seems to be the [putative name of a standardized version of `posix_spawn_file_actions_addchdir_np`](https://github.com/kraj/musl/commit/74244e5b3ed4a61d99c5fc0967b69e5c9a753456). Even when it's added, the only symbol is not going away.

So the question is what to do for versions of glibc that don't include `posix_spawn_file_actions_addchdir_np`. In that case you need to provide an actual implementation for glibc (perhaps copying from the glibc sources instead; the musl implementation at least is simple enough). 

You need to make sure, however, that your glibc-target implementation isn't run on musl and optionally that it doesn't run if glibc already provides the function. For that, you need to use `dlsym`. So:

* attempt to load libc.so.6. if it succeeds we're presumably in glibc. Now use dlsym to load `posix_spawn_file_actions_addchdir_np`. if it succeeds, we cache the value and invoke it. If not, we put a sentinel value in the cache, and call our own implementation
* attempt to load libc.musl-XXX.so.1. Use dlsym. If it succeeds, cache and call. O/wise abort() (unsupported/old musl)

cataphract (changes requested):

build_env/glibc_compat.c
cataphract: Actually, looking more closely, this isn't as simple. This is because `posix_spawn` needs to be changed as well to account for the operation in `posix_spawn_file_actions_addchdir_np`. It's still _doable_, though I'm not sure if would actually be simpler to use gnulib to replace all the spawn.h functions. This would possibly also imply changing spawn.h during compilation, since the structures `posix_spawnattr_t` and `posix_spawn_file_actions_t` are not opaque.

The simplest option though, would be to avoid using this function altogether...

dmehala: Good catch thanks. I completely missed that `posix_spawn` would needs to be aware and handle the actions set by `posix_spawn_file_actions_addchdir_np`. I'll give more thought into it.

PR #171, opened 2025-02-19
Block based on response status or headers

cataphract:
Significantly changes how the hooking mechanism for executing the final WAF run. The purpose is to allow blocking based on the response headers and status that would otherwise be sent.

Previously, we would install only an output filter. By the time it was invoked, the headers would already have been committed. It would not be possible to fully replace the response with another one.

So this PR installs also a header filter. When invoked, this header filter prevents the final filter from committing the response by changing the `send_chain` function pointer, continuing the filter chain and saving the buffers instead (because the filter may not invoke send_chain, it needs also to look at where this data may be buffered: this needs different code for http 1/3 and http/2).

Afterwards, it launches the WAF task. While this task is running, any data send to the output buffer is consumed as saved (copied), up to a certain limit. Once this limit is reached, the buffers are not consumed anymore, therefore stalling the output chain with busy buffers.

Once the final WAF task finishes, either a blocking response is committed, fully replacing the headers and body, or, if we're not to block, the buffered headers and body are sent forward (with special logic wrt the headers in http/2, as ngx_http_write_filter() can't send header frames in http/2).

Other changes:

* Blocking tests in http/2 and /3
* Fix read out-of-bounds in ngx_header_writer.h


dmehala:
@cataphract We're not part of the same team or even the same organization. Given that, providing some context upfront can help set the stage for those reviewing your PR. A bit of background on what it does and why it’s needed would go a long way
in ensuring a smoother review process. Could you, please, share more details?

cataphract:
> @cataphract We're not part of the same team or even the same organization. Given that, providing some context upfront can help set the stage for those reviewing your PR. A bit of background on what it does and why it’s needed would go a long way in ensuring a smoother review process. Could you, please, share more details?

I've added a description now. It was meant to be a draft PR; I opened it to run the pipelines.

src/security/context.cpp
```diff
@@ -45,6 +50,35 @@ namespace {
...
+std::size_t chain_size(ngx_chain_t const *ch) {
+  std::size_t size = 0;
+  for (ngx_chain_t const *cl = ch; cl; cl = cl->next) {
+    size += ngx_buf_size(cl->buf);
+  }
+  return size;
+}
+std::size_t has_special(ngx_chain_t const *ch) {
+  for (ngx_chain_t const *cl = ch; cl; cl = cl->next) {
+    return ngx_buf_special(ch->buf);
```
Anilm3, line 69: This doesn't look correct as you're returning on the first iteration and you're not even looking at `cl`.

src/security/context.cpp
```diff
@@ -45,6 +50,35 @@ namespace {
...
+  }
+  return false;
+}
+std::size_t has_last(ngx_chain_t const *ch) {
+  for (ngx_chain_t const *cl = ch; cl; cl = cl->next) {
+    if (ch->buf->last) {
```
Anilm3, line 75: Same here

src/security/context.cpp
```diff
@@ -255,7 +295,9 @@ class PolTaskCtx {
     try {
       ngx_log_debug(NGX_LOG_DEBUG_HTTP, req_log(), 0, "before task %p main",
                     this);
-      block_spec_ = static_cast<Self *>(this)->do_handle(*tp_log);
+      block_spec_ = as_self().do_handle(*tp_log);
+      // test long libddwaf call
+      // ::usleep(2000000);
```
Anilm3, line 300: Stray comments?

src/security/context.cpp
```diff
@@ -870,94 +1008,524 @@ ngx_int_t Context::do_request_body_filter(ngx_http_request_t &request,
...
+            if (!new_buf->pos) {
+              return NGX_ERROR;
+            }
+            new_buf->last = ngx_copy(new_buf->pos, buf->pos, size);
+            buf->pos = buf->last;  // consume
+            filter_ctx.copied_total += size;
+          }
+        }
+
+        size = new_buf->file_last - new_buf->file_pos;
```
Anilm3, line 1100: If `buf->temporary` would size be any different than what was already calculated?

src/security/context.cpp
```diff
@@ -870,94 +1008,524 @@ ngx_int_t Context::do_request_body_filter(ngx_http_request_t &request,
...
+    }
+
+    current_ctx_ = nullptr;
+    request.connection->send_chain = prev_send_chain_;
+  }
+
+ private:
+  static ngx_chain_t *send_chain_save(ngx_connection_t *c, ngx_chain_t *in,
+                                      off_t limit) {
+    (void)limit;
```
Anilm3, line 1246: You can use `[[maybe_unused]]` instead.

src/security/context.cpp
```diff
@@ -870,94 +1008,524 @@ ngx_int_t Context::do_request_body_filter(ngx_http_request_t &request,
...
+
+    ngx_http_request_t *req = static_cast<decltype(req)>(c->data);
+    assert(req != nullptr);
+
+    auto *ctx = instance.current_ctx_;
+    assert(ctx != nullptr);
```
Anilm3, line 1252: How certain are we about these assertions...? An explicit check may be more prudent otherwise.

Anilm3 (approved):
From reading the PR it's quite hard to ascertain if this will work as intended, at least not without spending a similar amount of time it took to write it reviewing it... However aside from a few comments, I don't see any critical issues although some mistakes have been highlighted and should be fixed.

src/security/context.cpp
cataphract: I think mixed file/temporary buffers are the same size; at least that's what I saw. So yes, this line can be put in an else. (At the end, it doesn't really matter either way, as this size is only used for logging -- unlike `copied_total`)

src/security/context.cpp
cataphract: What's not clear is that c->data is the request in http3.

Seeing as this function is very difficult to exercise and the tests don't reach it (the data for the headers is small enough that it isn't sent just yet), I'm changing this to not rely on c->data.

src/security/context.cpp
cataphract: I just left marked where I should uncomment to test long libddwaf calls because it's not an easy thing to test otherwise.

src/security/context.h
```diff
@@ -27,6 +24,8 @@ extern "C" {
 
 namespace datadog::nginx::security {
 
+static constexpr uintptr_t kBufferTag = 0xD47AD06;
```
dmehala, line 27: [suggestion] Can you add a comment to explain this magic number, its purpose and how it is used?

dmehala (approved):

src/security/context.h
cataphract: it's an arbitrary tag to use for our buffers. modules usually use the value of the pointer to the module structure, but this has the advantage it's available at compile time. It's "datadog" written in hex digits

PR #175, opened 2025-02-26
Upgrade to libddwaf 1.24.0

cataphract:


Anilm3:
Minor annoyance, but we might change the fallback on ASM_DD as we're considering supporting multiple configurations (check the RFC for the comments). I'll try to address that ASAP so it doesn't block you.

Anilm3 (approved):
lgtm

PR #176, opened 2025-03-03
chore: remove unecessary diagnostics

dmehala:
Preparatory work for issue#173


src/datadog_conf.cpp
```diff
@@ -1,5 +1,7 @@
 #include "datadog_conf.h"
 
+#include <format>
```
dubloom, line 3: nit: did you forget that ?

dubloom (approved):
Just a question to understand, did we have any advantage to store the location in the configured_value before ?

pablomartinezbernardo (approved):

PR #177, opened 2025-03-03
feat: evaluate `datadog_*` value during request processing

### LINKED ISSUE #173: [FEATURE]: datadog_service_name NGINX variables support
### Describe the goal of the feature

It would be great if the `datadog_service_name` directive supported [NGINX variables](https://nginx.org/en/docs/varindex.html). With nginx-datadog on Ingress-NGINX, this would enable dynamically assigning a service name to each ingress object.

```
controller:
  config:
    main-snippet: "load_module /modules_mount/ngx_http_datadog_module.so;"
    http-snippet: |
      server {
        datadog_sample_rate 0.1;
      }

    location-snippet: |
      datadog_service_name "$ingress_name-ingress";
```

### Is your feature request related to a problem?

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

We are looking to transition from our current setup, which uses the OpenTelemetry module (and previously OpenTracing) bundled in Ingress-NGINX, to the nginx-datadog module for improved support. Here is our current configuration:

OpenTelemetry
```
location-snippet: |
  opentracing_tag service.name "$ingress_name-ingress";
```

OpenTracing
```
location-snippet: |
  opentelemetry_attribute "service.name" "$ingress_name-ingress";

### PR DESCRIPTION
**Pull request by dmehala**
Changes:
  - Replace `NgxScript` with `ngx_http_complex_value`.
  - Use default NGINX conf setter for complex value.

Resolves #173

### DIFF
--- a/src/ngx_http_datadog_module.cpp
+++ b/src/ngx_http_datadog_module.cpp
@@ -162,32 +162,32 @@ static ngx_command_t datadog_commands[] = {
       "datadog_operation_name",
       "opentracing_operation_name",
       anywhere | NGX_CONF_TAKE1,
-      set_datadog_operation_name,
+      ngx_http_set_complex_value_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
-      0,
+      offsetof(datadog_loc_conf_t, operation_name_script),
       nullptr),
 
     DEFINE_COMMAND_WITH_OLD_ALIAS(
       "datadog_location_operation_name",
       "opentracing_location_operation_name",
       anywhere | NGX_CONF_TAKE1,
-      set_datadog_location_operation_name,
+      ngx_http_set_complex_value_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
-      0,
+      offsetof(datadog_loc_conf_t, loc_operation_name_script),
       nullptr),
 
     { ngx_string("datadog_resource_name"),
       anywhere | NGX_CONF_TAKE1,
-      set_datadog_resource_name,
+      ngx_http_set_complex_value_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
-      0,
+      offsetof(datadog_loc_conf_t, resource_name_script),
       nullptr},
 
     { ngx_string("datadog_location_resource_name"),
       anywhere | NGX_CONF_TAKE1,
-      set_datadog_location_resource_name,
+      ngx_http_set_complex_value_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
-      0,
+      offsetof(datadog_loc_conf_t, loc_resource_name_script),
       nullptr},
 
     DEFINE_COMMAND_WITH_OLD_ALIAS(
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by pablomartinezbernardo — APPROVED** [review]

File: src/ngx_http_datadog_module.cpp [hunk 1]
```diff
@@ -162,32 +162,32 @@ static ngx_command_t datadog_commands[] = {
...
+      offsetof(datadog_loc_conf_t, resource_name_script),
       nullptr},
 
     { ngx_string("datadog_location_resource_name"),
       anywhere | NGX_CONF_TAKE1,
-      set_datadog_location_resource_name,
+      ngx_http_set_complex_value_slot,
       NGX_HTTP_LOC_CONF_OFFSET,
-      0,
+      offsetof(datadog_loc_conf_t, loc_resource_name_script),
```
dubloom, line 190: Is there an interest having two directives if at the end it is the same used function ?

src/ngx_http_datadog_module.cpp
```diff
@@ -709,40 +709,31 @@ static void *create_datadog_loc_conf(ngx_conf_t *conf) noexcept {
     return nullptr;  // error
   }
 
-#ifdef WITH_RUM
-  loc_conf->rum_enable = NGX_CONF_UNSET;
-  loc_conf->rum_snippet = nullptr;
```
dubloom, line 714: Why ? Because it is useless (set null by default I guess) ?

dubloom (commented):
May leave other comments after lunch

src/ngx_http_datadog_module.cpp
dmehala: To be coherent, I now use class member initialization. Check src/datadog_conf.h.

src/ngx_http_datadog_module.cpp
dubloom: If you look at the "datadog_resource_name" and "datadog_location_resource_name", everything is similar except the directive name. Changing that might introduces a breaking change so it does not have to be in this PR if it is the case but one of them looks useless.

src/request_tracing.cpp
```diff
@@ -38,42 +38,45 @@ bool should_delegate(ngx_http_request_t *request,
   return loc_conf->sampling_delegation_enabled &&
          loc_conf->allow_sampling_delegation_in_subrequests;
 }
+
+std::optional<std::string> set_if_exists(ngx_http_complex_value_t *conf,
```
dubloom, line 42: nit: this is not setting anything so i would change the name of the function

dubloom (approved):
except the two small comments, LGTM

src/ngx_http_datadog_module.cpp
dmehala: `datadog_resource_name` set the `resource_name_script` variable, while `datadog_location_resource_name` set `loc_resource_name_script`. These are two _different_ variables,

src/request_tracing.cpp
dmehala: addressed in 248a4c32bc084b3b955c31baa5bcb43dfa27c295

src/ngx_http_datadog_module.cpp
dubloom: Well

dubloom: I read that part too quickly

dubloom: my bad

PR #180, opened 2025-03-06
chore: default variables IDs to 128-bit

dmehala:
# Description
Since https://github.com/DataDog/nginx-datadog/pull/103 two versions of the module has been released. It is finally the time to default `datadog_*_id` variables to 128-bit IDs by default.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -456,11 +456,11 @@ Note that if `datadog_trace_locations` is `on`, then `$datadog_span_id` will
 refer to the span associated with the location (outbound request), not its
 parent (inbound request).
 
-### `datadog_trace_id_dec`  
+### `datadog_trace_id_64bits_base10`  
 This reflects the legacy behavior prior to `v1.6.0`, where `$datadog_trace_id` evaluated to the decimal representation of the 64-bit trace ID.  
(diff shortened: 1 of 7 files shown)


doc/API.md
```diff
@@ -443,24 +443,26 @@ The Datadog nginx module defines additional variables that provide information
...
 ID of the currently active span.If there is no currently active span, then
 the variable expands to a hyphen character (`-`) instead.
 
 Note that if `datadog_trace_locations` is `on`, then `$datadog_span_id` will
 refer to the span associated with the location (outbound request), not its
 parent (inbound request).
 
-### `datadog_trace_id_hex`
-Same as `$datadog_trace_id`, but is the hexadecimal representation of the 128-bit ID. 
+### `datadog_trace_id_dec`
```
zacharycmontoya, line 459: Since naming is a fun engineering problem 😄 WDYT about `_base10` instead of `_dec`? Not blocking, just a suggestion

zacharycmontoya (approved):

doc/API.md
dmehala: Why not. how about `_short_base10`.

zacharycmontoya: Sounds good to me!

PR #181, opened 2025-03-06
chore: relocate products directives to submodule

dmehala:
This change involves moving the directive definitions to their respective "submodules". This is an initial step toward modularizing the module, ensuring a clear separation of concerns between the different products.

Preparatory work for drop-in support for opentelemetry.


src/datadog_directive.cpp
```diff
@@ -1,380 +1,9 @@
...
-
-  ngx_log_error(
-      NGX_LOG_WARN, cf->log, 0,
-      "Directive \"%V\" is deprecated and can be removed since v1.2.0.",
-      &elements[0]);
-
-  return NGX_OK;
-}
 
 char *warn_deprecated_command_datadog_tracing(ngx_conf_t *cf,
```
dubloom, line 8: Why are we letting that here and not in tracing/directives ?

src/rum/config.h
```diff
@@ -22,17 +23,24 @@ char *datadog_rum_merge_loc_config(ngx_conf_t *cf,
...
-    { \
-      ngx_string("datadog_rum"), \
-      NGX_HTTP_MAIN_CONF | NGX_HTTP_SRV_CONF | NGX_HTTP_LOC_CONF | NGX_CONF_FLAG, \
-      ngx_conf_set_flag_slot, NGX_HTTP_LOC_CONF_OFFSET, offsetof(datadog_loc_conf_t, rum_enable), NULL \
-    }, \
-    { \
-      ngx_string("datadog_rum_config"), \
-      NGX_HTTP_MAIN_CONF | NGX_HTTP_SRV_CONF | NGX_HTTP_LOC_CONF | NGX_CONF_BLOCK | NGX_CONF_TAKE1, \
-      on_datadog_rum_config, NGX_HTTP_LOC_CONF_OFFSET, 0, NULL \
+const datadog::nginx::directive rum_directives[] = {
```
dubloom, line 26: Shouldn't we put that in a directives.cpp file ? I can do it in another PR.

src/datadog_directive.h
```diff
@@ -7,47 +7,76 @@ extern "C" {
...
-                      void *conf) noexcept;
+  ngx_command_t to_ngx_command() const {
+    return ngx_command_t{to_ngx_str(name), type, callback, conf, offset, post};
+  }
+};
 
-char *json_config_deprecated(ngx_conf_t *cf, ngx_command_t *command,
-                             void *conf) noexcept;
+template <std::size_t N1, std::size_t... Ns>
+constexpr auto merge_directives(const directive (&arr1)[N1],
```
dubloom, line 49: I trust you on this function but maybe it could be a good idea to add some comments as it is a bit an obscure function (from my point of view at least).

dubloom (approved):
Overall LGTM. However i'd delete the datadog_directive.cpp file and put the directive in tracing/directives as it is a tracing related directive.
It good be nice to create a directives.h in rum folder to be consistent

src/datadog_directive.cpp
dmehala: It could be used regardless of the product (rum, appsec or tracing). Think about it as a common function to handle deprecated directives.

src/datadog_directive.h
dmehala: Well, it's metaprogramming. If the reader has no knowledge of metaprogramming, it's sort of expected it seems a bit obscure.

src/datadog_directive.cpp
dmehala: Good catch, that's what I did in otel PR.

PR #182, opened 2025-03-07
feat: otel drop in support

dmehala:
# Description
This PR introduces the ability to replace the OpenTelemetry (oTel) module`otel_ngx_module` with `nginx-datadog` without requiring any changes to the existing oTel directives. This enhancement aims to simplify the migration process for ingress-nginx users, enabling them to switch from the `otel_ngx_module` to `nginx-datadog` or the opposite seamlessly, without the need to modify their kubernetes CRD/resources.

### Changes
- Introduced aliases for compatible `otel_ngx_module` directives, ensuring they are ignored when no corresponding `nginx-datadog` equivalent exists.
- Added aliases for compatible `otel_ngx_module` variables to match their `nginx-datadog` equivalents.
- Included several helper directives.
- Replaced instances of `ngx_log_error` with `ngx_conf_log_error` where applicable. This change simplifies the code by leveraging `ngx_conf_log_error`'s ability to automatically print the file and location of the error in the configuration file.

### DIFF
--- a/src/datadog_directive.cpp
+++ b/src/datadog_directive.cpp
@@ -2,21 +2,73 @@
 
 #include <cassert>
 
+#include "datadog_conf_handler.h"
+
 namespace datadog {
 namespace nginx {
 
-char *warn_deprecated_command_datadog_tracing(ngx_conf_t *cf,
-                                              ngx_command_t * /*command*/,
-                                              void * /*conf*/) noexcept {
+char *silently_ignore_command(ngx_conf_t *cf, ngx_command_t *command, void *) {
+  ngx_conf_log_error(NGX_LOG_DEBUG, cf, 0, "Directive \"%V\" ignored",
+                     &command->name);
+  return NGX_OK;
+}
+
+char *alias_directive(ngx_conf_t *cf, ngx_command_t *command, void *) noexcept {
+  if (command->post == nullptr) {
+    return static_cast<char *>(NGX_CONF_ERROR);
+  }
+
+  auto src_directive = static_cast<char *>(command->post);
+
   const auto elements = static_cast<ngx_str_t *>(cf->args->elts);
   assert(cf->args->nelts >= 1);
 
-  ngx_log_error(
-      NGX_LOG_WARN, cf->log, 0,
-      "Directive \"%V\" is deprecated. Use datadog_tracing on/off instead",
-      &elements[0]);
+  const ngx_str_t new_name_ngx{.len = strlen(src_directive),
+                               .data = (u_char *)src_directive};
+  ngx_conf_log_error(NGX_LOG_DEBUG, cf, 0, "Alias \"%V\" to \"%V\"",
+                     &command->name, &new_name_ngx);
 
+  // Rename the command and let `datadog_conf_handler` dispatch to the
+  // appropriate handler.
+  elements[0] = new_name_ngx;
+  auto rcode = datadog_conf_handler({.conf = cf, .skip_this_module = false});
+  if (rcode != NGX_OK) {
+    return static_cast<char *>(NGX_CONF_ERROR);
+  }
+
+  return static_cast<char *>(NGX_CONF_OK);
+}
+
+char *warn_deprecated_command(ngx_conf_t *cf, ngx_command_t *command,
+                              void *) noexcept {
+  u_char buf[NGX_MAX_ERROR_STR] = {0};
+  u_char *last = buf + NGX_MAX_ERROR_STR;
+  u_char *p =
+      ngx_slprintf(buf, last, "Directive \"%V\" is deprecated", &command->name);
+
+  if (command->post != nullptr) {
+    auto reason = static_cast<char *>(command->post);
+    ngx_slprintf(p, last, ". %s", reason);
+  }
+
+  ngx_conf_log_error(NGX_LOG_WARN, cf, 0, "%s", buf);
   return NGX_OK;
+}
+
+char *err_deprecated_command(ngx_conf_t *cf, ngx_command_t *command,
+                             void *) noexcept {
+  u_char buf[NGX_MAX_ERROR_STR] = {0};
+  u_char *last = buf + NGX_MAX_ERROR_STR;
+  u_char *p =
+      ngx_slprintf(buf, last, "Directive \"%V\" is deprecated", &command->name);
+
+  if (command->post != nullptr) {
+    auto reason = static_cast<char *>(command->post);
+    ngx_slprintf(p, last, ". %s", reason);
+  }
+
+  ngx_conf_log_error(NGX_LOG_WARN, cf, 0, "%s", buf);
+  return static_cast<char *>(NGX_CONF_ERROR);
 }
 
 }  // namespace nginx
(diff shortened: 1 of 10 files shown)


test/cases/orchestration.py
dmehala, line 1: - [ ] revert

src/datadog_directive.cpp
```diff
@@ -2,22 +2,74 @@
...
+#include "datadog_conf_handler.h"
+
 namespace datadog {
 namespace nginx {
 
-char *warn_deprecated_command_datadog_tracing(ngx_conf_t *cf,
-                                              ngx_command_t * /*command*/,
-                                              void * /*conf*/) noexcept {
+char *silently_ignore_command(ngx_conf_t *cf, ngx_command_t *command, void *) {
+  ngx_conf_log_error(NGX_LOG_DEBUG, cf, 0, "Directive \"%V\" ignored",
```
dubloom, line 11: It is silent in terms of logging the error but let the program continue, right ?

src/datadog_directive.cpp
```diff
@@ -2,22 +2,74 @@
...
+                     &command->name);
+  return NGX_OK;
+}
+
+char *alias_directive(ngx_conf_t *cf, ngx_command_t *command, void *) noexcept {
```
dubloom, line 16: I have a few comments on this function:
- Can't we replace the assertion by a return of NGX_CONF_ERROR ? Because at the moment it would mean that the program is just crashing, no ?
- Do we really want to log the alias every times ?

src/datadog_variable.cpp
```diff
@@ -52,16 +48,29 @@ static ngx_int_t expand_span_variable(ngx_http_request_t* request,
...
+    variable_value->not_found = false;
+    variable_value->data = span_variable_value.data;
+  };
+
+  auto variable_name = to_string_view(*reinterpret_cast<ngx_str_t *>(data));
+  if (variable_name.starts_with("datadog_")) {
+    auto suffix =
+        slice(variable_name, TracingLibrary::span_variables().prefix.size());
+    set_variable(suffix);
+  } else if (std::string_view otel("opentelemetry_");
```
dubloom, line 66: Above we are not using a string view. Can we make the two cases consistent ?

dubloom: Ok, we can use prefix.size(). So it's a nit comment (I'm fine with staying like that)

pablomartinezbernardo (approved):
LGTM

src/datadog_directive.h
```diff
@@ -72,11 +72,39 @@ constexpr auto generate_directives(const T &...directives) {
   return merge_directives(directives...);
 }
 
+#define IGNORE_COMMAND(NAME, TYPE) \
+  { NAME, TYPE, silently_ignore_command, NGX_HTTP_LOC_CONF_OFFSET, 0, nullptr }
+
+#define ALIAS_COMMAND(SRC_DIR, TARGET_DIR, TYPE)                    \
```
dubloom, line 78: For me, src_dir evokes src_directory so I was a bit lost. To improve readability, I would argue that it would be better to fully write SRC_DIRECTIVE.

It is a bit more of a nit comment but similarly, when I see ALIAS_COMMAND("datadog_trust_incoming_span", "opentracing_trust_incoming_span"...), it evokes me that when we use datadog_trust we want to use opentracing while it is more the reverse

src/ngx_http_datadog_module.cpp
```diff
@@ -52,18 +52,13 @@ static char *merge_datadog_loc_conf(ngx_conf_t *, void *parent, void *child) noe
...
-#define DEFINE_DEPRECATED_COMMAND_DATADOG_TRACING(NAME, TYPE) \
-  {                                                           \
-    NAME, TYPE, warn_deprecated_command_datadog_tracing,      \
-        NGX_HTTP_LOC_CONF_OFFSET, 0, NULL                     \
-  }
-
 constexpr datadog::nginx::directive module_directives[] = {
-    DEFINE_DEPRECATED_COMMAND_DATADOG_TRACING("datadog_enable",
-                                              anywhere | NGX_CONF_NOARGS),
+    WARN_DEPRECATED_COMMAND("datadog_enable", anywhere | NGX_CONF_NOARGS,
```
dubloom, line 56: Shouldn't we move this to tracing/directives ?

test/cases/variables/conf/otel_drop_in_support.conf
```diff
@@ -0,0 +1,27 @@
+# "/datadog-tests" is a directory created by the docker build
+# of the nginx test image. It contains the module, the
+# nginx config, and "index.html".
```
dubloom, line 3: I'm a bit lost on how this is testing otel ? We add aliases for a bunch of otel directives but we are not really testing them.

dubloom (changes requested):
The code LGMT except a slight potential readability enhancement. However, it seems that the aliases for otel directives are not testsed

src/datadog_directive.cpp
dubloom: Auto answer: if the DEBUG_LOG are something we need to activate you can ignore my second question

src/datadog_directive.cpp
dubloom: Auto answer: if the DEBUG_LOG are something we need to activate you can ignore this

src/ngx_http_datadog_module.cpp
dmehala: No, they are globally applied directives.

test/cases/variables/conf/otel_drop_in_support.conf
dmehala: Quite simple. It uses `$opentelemetry_trace_id` and `$opentelemetry_span_id` variable. Those variables are aliases on `$datadog_trace_id` and `$datadog_span_id` respectively. This test ensures the new `$opentelemetry_` variables are defined and returns the expected value.

src/datadog_directive.cpp
dmehala: > Can't we replace the assertion by a return of NGX_CONF_ERROR ?

That's an interesting take. I do believe assert should be enabled in release build but that's not the case at the moment. Assertion in general are very unlikely scenarios that you mean to catch during development.

Here, it is _very_ unlikely because the NGINX API guarantee that the number of element is X if the directive is correctly defined. If this invariant is false, either the directive definition is wrong and will be catch during testing OR NGINX do not behave as expected.

> Do we really want to log the alias every times ?
It's a DEBUG_LOG, by essence, debug logs are quite verbose. Consequently, this is a non-issue.

src/datadog_directive.h
dmehala: Addressed in 79ebeff405491b385fce2e63b4b35ab0b409d58e.

> when I see ALIAS_COMMAND("datadog_trust_incoming_span", "opentracing_trust_incoming_span"...), it evokes me that when we use datadog_trust we want to use opentracing while it is more the reverse

This is a subjective take. At the end of the day, somebody will not be happy about. I guess I could use `constexpr` function with designated initializers. If that can help, it uses the same order than all posix command (`mv`, `cp` ...)

src/datadog_directive.cpp
dmehala: ditto

test/cases/variables/conf/otel_drop_in_support.conf
dubloom: Yes but we are not testing directive alias

dmehala: Oh that, indeed, good catch.

dmehala: I will add tests in another PR. I want to release a new version at best next week.

PR #183, opened 2025-03-11
feat!: Allow overriding default tags

### LINKED ISSUE #172: [FEATURE]: Ability to hide queryString from generated spans
### Describe the goal of the feature

When we switched from an old version of _ingress-nginx_ instrumented with _nginx-opentracing_ module to using _nginx-datadog_ , we noticed query parameters are appended to the `http.url` span attribute. Also, there is a new span attribute `http.url_details.queryString` that contains the list of all query parameter.

I am not sure if this is due to the switch to _nginx-datadog_ though. I couldn't find a related difference in code between _nginx-opentracing_ and _nginx-datadog_.

Regardless, an ability to hide query string from generated spans would be useful. There could be services that write PII data on query string, even though that is a bad security practice. An ability to hide it will avoid that from surfacing on Datadog.

### Is your feature request related to a problem?

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

dmehala:
Introduces the ability to override default tags using `datadog_set`.

Changes:
- Refactor how tags are stored.
- BREAKING CHANGE: only value tags can contain expressions which can be evaluated during request processing.
- Update directive callbacks to use `NGX_CONF_OK` instead of `NGX_OK` for better consistency.

Resolves #172 

## How this solves #172?
This allows `http.url` to be overriden. To remove query string from the URL, one can set `http.url` to`$scheme://$http_host$uri`, where `$uri` is the normalized version of `$request_uri` used by default.

Usage example:
```
http {
  datadog_tag "http.url" "$scheme://$http_host$uri";
  
  location / {
    return 200 "hello, worl";
  }
}
```


test/cases/tags/test_tags.py
```diff
@@ -91,39 +91,81 @@ def test_default_tags(self):
...
+    def test_overwite_default_tags(self):
+        """Verify default tags can be overwritten"""
+        conf_path = Path(__file__).parent / "./conf/overwrite_defaults.conf"
+        conf_text = conf_path.read_text()
+        self.orch.nginx_replace_config(conf_text, conf_path.name)
+
+        # Consume any previous logging from the agent.
+        self.orch.sync_service("agent")
+
+        status, _, _ = self.orch.send_nginx_http_request("/http")
```
dmehala, line 142:
```suggestion
        status, _, _ = self.orch.send_nginx_http_request("/http?key=value")
```

src/tracing/directives.cpp
```diff
@@ -41,31 +41,45 @@ char *lock_propagation_styles(const ngx_command_t *command, ngx_conf_t *conf) {
...
+  ccv.complex_value = cv;
+  ccv.zero = 0;
+  ccv.conf_prefix = 0;
+
+  if (ngx_http_compile_complex_value(&ccv) != NGX_OK) {
+    return nullptr;
+  }
 
-  return static_cast<char *>(NGX_CONF_OK);
+  return cv;
```
pablomartinezbernardo, line 62: Is it possible to reuse the implementation from common? Or does it actually need to be duplicated for some reason?

pablomartinezbernardo (approved):
Couple small comments

src/tracing/directives.cpp
dmehala: addressed in 655627098ce1667901127c2727c851e0cbe5a947

src/common/directives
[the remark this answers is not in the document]
dmehala: addressed in 655627098ce1667901127c2727c851e0cbe5a947

pablomartinezbernardo (approved):

PR #187, opened 2025-03-31
feat: report integration errors

dmehala:
Resolves [AIDM-394]

[AIDM-394]: https://datadoghq.atlassian.net/browse/AIDM-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/datadog_handler.cpp
+++ b/src/datadog_handler.cpp
@@ -1,14 +1,12 @@
 #include "datadog_handler.h"
 
+#include <datadog/telemetry/telemetry.h>
+
+#include "datadog_context.h"
 #include "ngx_http_datadog_module.h"
 
 extern "C" {
 #include <ngx_config.h>
-}
-
-#include "datadog_context.h"
-
-extern "C" {
 extern ngx_module_t ngx_http_datadog_module;
 }
 
@@ -31,6 +29,8 @@ static bool is_datadog_tracing_enabled(
     return loc_conf->enable_tracing && core_loc_conf->log_subrequest;
   }
 }
+#define CURRENT_FRAME \
+  std::format("{}:{} in {}", __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 ngx_int_t on_enter_block(ngx_http_request_t *request) noexcept try {
   auto core_loc_conf = static_cast<ngx_http_core_loc_conf_t *>(
@@ -53,7 +53,8 @@ ngx_int_t on_enter_block(ngx_http_request_t *request) noexcept try {
   } else {
     try {
       context->on_change_block(request, core_loc_conf, loc_conf);
-    } catch (...) {
+    } catch (const std::exception &e) {
+      telemetry::report_error_log(e.what());
       // The DatadogContext may be broken, destroy it so that we don't
       // attempt to continue tracing.
       destroy_datadog_context(request);
@@ -62,6 +63,7 @@ ngx_int_t on_enter_block(ngx_http_request_t *request) noexcept try {
   }
   return NGX_DECLINED;
 } catch (const std::exception &e) {
+  telemetry::report_error_log(e.what());
   ngx_log_error(NGX_LOG_ERR, request->connection->log, 0,
                 "Datadog instrumentation failed for request %p: %s", request,
                 e.what());
@@ -83,7 +85,8 @@ ngx_int_t on_access(ngx_http_request_t *request) noexcept try {
     return NGX_AGAIN;
   }
   return NGX_DECLINED;
-} catch (std::exception &e) {
+} catch (const std::exception &e) {
+  telemetry::report_error_log(e.what());
   ngx_log_error(NGX_LOG_ERR, request->connection->log, 0,
                 "Datadog instrumentation failed for request %p: %s", request,
                 e.what());
@@ -97,6 +100,7 @@ ngx_int_t on_log_request(ngx_http_request_t *request) noexcept {
   try {
     context->on_log_request(request);
   } catch (const std::exception &e) {
+    telemetry::report_error_log(e.what());
     ngx_log_error(NGX_LOG_ERR, request->connection->log, 0,
                   "Datadog instrumentation failed for request %p: %s", request,
                   e.what());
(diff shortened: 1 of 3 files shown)


pablomartinezbernardo (commented):
Do you know why tests might be failing?

cataphract:
On amazonlinux:2.0.20230418.0 there's a failure:

AssertionError: 1 != 0 : ['nginx: [emerg] dlopen() "/datadog-tests/ngx_http_datadog_module.so" failed (/datadog-tests/ngx_http_datadog_module.so: undefined symbol: memfd_create) in /tmp/fastcgi_auto.conf:5', 'nginx: configuration file /tmp/fastcgi_auto.conf test failed', '']

memfd_create was added in glibc 2.27 (and linux 3.17), but that image uses 2.26... You'll need to either provide the function (using `syscall`) (see a table with the number for [amd64/arm64](https://syscalls.mebeim.net/?table=x86/64/x64/latest)) or remove that call.

src/datadog_handler.cpp
```diff
@@ -53,7 +54,8 @@ ngx_int_t on_enter_block(ngx_http_request_t *request) noexcept try {
   } else {
     try {
       context->on_change_block(request, core_loc_conf, loc_conf);
-    } catch (...) {
+    } catch (const std::exception &e) {
+      telemetry::report_error_log(e.what(), CURRENT_FRAME(request));
```
dubloom, line 58: Using directly e.what() are we sure we are not taking any risks or leaking PII ?

src/telemetry_util.h
```diff
@@ -0,0 +1,38 @@
...
+  size_t pos = absolute_filepath.find(prefix);
+  return pos != std::string_view::npos ? absolute_filepath.substr(pos)
+                                       : absolute_filepath;
+}
+
+inline std::string make_current_frame(ngx_http_request_t *request,
+                                      std::string_view file, size_t line,
+                                      std::string_view function) {
+  // clang-format off
+  return std::format("Exception catched:\n"
```
dubloom, line 26:
```suggestion
  return std::format("Exception caught:\n"
```

src/telemetry_util.h
```diff
@@ -0,0 +1,38 @@
...
+}
+
+namespace datadog::nginx {
+
+constexpr std::string_view relative_filepath(
+    std::string_view absolute_filepath) {
+  constexpr std::string_view prefix = "nginx-datadog";
+  size_t pos = absolute_filepath.find(prefix);
+  return pos != std::string_view::npos ? absolute_filepath.substr(pos)
+                                       : absolute_filepath;
```
dubloom, line 19: The problem with this method is, if nginx-datadog is not in the string (which I agree with you should not happen), we will leak the full_patch, can we add a safe guard please ?

dmehala: Could you clarify what the potential risks are of exposing the full path? (which is _highly_ improbable unless the repo change name or the ci building the project is cloning the repo under another name). Also, could you elaborate on what you mean by safeguard? Do you have any suggestions or proposals in mind?

src/telemetry_util.h
dmehala: That's an interesting one. `catched` is actually valid English and commonly used before 19th century, which is when `caught` started to replaced it.

src/telemetry_util.h
dubloom: For instance, i have my name.surname in my full_path, which is PII. In PHP, I did this:
https://github.com/DataDog/dd-trace-php/blob/0e702530af821f358ccd1dbdecd9535728eed459/ext/telemetry.c#L50

I agree that is a bit overkill and I could accept to not include that it does not add runtime overhead so I think "prevention is better than cure" is the way to go.

src/telemetry_util.h
dubloom: Oh, so every time I said catched I was not making any mistake. Nice to know haha

src/datadog_handler.cpp
dubloom: PHP is in the same case as you so we are not handling them (and I will have to check this is safe). 
In python i'm ensuring the log message is constant. 

If you know that we are not throwing exceptions with PIIs we don't have to make any changes.

src/telemetry_util.h
dmehala: Interesting 🤔 Make sense. I'll adjust the method to return only the name of the file instead of the full path. Thanks for raising your concerns.

dmehala: addressed in e5d20039095d87d1e576fbd203f823d0c8af0357

src/telemetry_util.h
```diff
@@ -0,0 +1,41 @@
...
+
+namespace datadog::nginx {
+
+constexpr std::string_view relative_filepath(
+    std::string_view absolute_filepath) {
+  constexpr std::string_view prefix = "nginx-datadog";
+  const size_t pos = absolute_filepath.find(prefix);
+  if (pos != std::string_view::npos) return absolute_filepath.substr(pos);
+
+  const size_t sep_pos = absolute_filepath.find_first_of("/\\");
```
dubloom, line 20: nit: I think we should use `find_last_of` here. 
otherwise something like /Users/name.surname would still leak name.surname

dubloom (commented):
Are we interested in reporting [tracer initialization error](https://github.com/DataDog/nginx-datadog/blob/a615fa2e63aaea29a5645f38e4ade4c310890fe0/src/ngx_http_datadog_module.cpp#L368) ?

dubloom (commented):
Thanks for addressing my comment. 

I'd still have the two same additionnal comments:
- Are we ignoring the tracer initialization error by purpose ?
- Are we considering adding a test for the telemetry (if possible) ?

dmehala:
> Are we ignoring the tracer initialization error by purpose ?
Yes. If tracer initialization fails, it's mostly because the configuration is wrong. I don't see the value in reporting initialization failure based on users misconfiguration but I can be wrong.

> Are we considering adding a test for the telemetry (if possible) ?
No 'cause there is no reliable way to trigger these state.

dubloom (approved):
LGTM 😀

PR #189, opened 2025-04-03
fix(test): collect coverage without using bash

dmehala:
## Description
Recent versions of `nginx:<VERSION>-alpine` docker images does not have `bash` installed anymore making the collection of code coverage failing... and test failing. This remove the dependency on bash.

### DIFF
--- a/test/cases/orchestration.py
+++ b/test/cases/orchestration.py
@@ -3,6 +3,8 @@
 from . import formats
 from .lazy_singleton import LazySingleton
 
+import tarfile
+import tempfile
 import contextlib
 import faulthandler
 import json
@@ -480,7 +482,6 @@ class Orchestration:
         Run `docker compose down` to bring down the orchestrated services.
         Join the log-parsing thread.
         """
-
         if self.has_coverage_data():
             self.create_coverage_tarball()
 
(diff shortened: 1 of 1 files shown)


test/cases/orchestration.py
```diff
@@ -506,37 +507,45 @@ def down(self):
...
             "-T",
             "--",
             "nginx",
-            "bash",
-            "-c",
-            "tar --transform='s@tmp/@@' -czf - -T <(find /tmp -maxdepth 1 -name '*.profraw')",
+            "find",
+            "/tmp",
+            "-name",
+            "*.profraw",
```
cataphract, line 528: I think it would be simpler to just avoid process substitution and pass the files to tar on the command line:

```python
tar_command = docker_compose_command(
    "exec",
    "-T",
    "--",
    "nginx",
    "sh",
    "-c",
    "find /tmp -maxdepth 1 -name '*.profraw' | xargs tar --transform='s@tmp/@@' -czf -",
)
```

cataphract (approved):

test/cases/orchestration.py
dmehala: That's a possibility. It's clever, however, I am afraid not everyone can understand this command line, whereas python code is much more expressive.

PR #197, opened 2025-04-16
fix(test): rework `reload_nginx`

dmehala:
This change comes after a deep investigation into why certain jobs were failing when upgrading to the latest `dd-trace-cpp` version. The root cause turned out to be increased contention in the default curl HTTP client due to telemetry refactoring. This caused NGINX to take significantly longer to shut down, which in turn led to timeouts during `reload_nginx`.

~~While investigating, I also realized the logic in `reload_nginx` was flawed. It was only checking whether a worker process was running, rather than verifying that the old worker (by PID) had exited. Since NGINX spawns a new worker and sends a SIGQUIT to the old one during reloads, our previous check was resulting in false positives. I've updated the logic to explicitly wait for the old worker PID to terminate and~~ confirm the new worker is up before proceeding.

Additionally, while fixing this, I encountered issues with some AppSec tests. These tests were intentionally causing the worker to fail during initialization (e.g. by using invalid config paths), but this left NGINX in a broken state for subsequent tests. To resolve this, the tests for
`datadog_appsec_http_blocked_template_json`,
`datadog_appsec_http_blocked_template_html`, and
`datadog_appsec_ruleset_file` now validate the existence of their required files during the configuration process instead of at worker startup.

### DIFF
--- a/test/cases/orchestration.py
+++ b/test/cases/orchestration.py
@@ -29,6 +29,21 @@ def quit_signal_handler(signum, frame):
     faulthandler.dump_traceback()
 
 
+def wait_until(predicate_func, timeout_seconds):
+    """Wait until `predicate_func` is True."""
+    before = time.monotonic()
+    while True:
+        if predicate_func():
+            break
+
+        now = time.monotonic()
+        if now - before >= timeout_seconds:
+            raise Exception(
+                f"{timeout_seconds} seconds timeout exceeded while waiting for nginx workers to stop.  {now - before} seconds elapsed."
+            )
+        time.sleep(0.5)
+
+
 signal.signal(signal.SIGQUIT, quit_signal_handler)
 
 # Since we override the environment variables of child processes,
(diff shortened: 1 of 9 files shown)


test/cases/sec_config/test_sec_config.py
```diff
@@ -1,5 +1,6 @@
 import json
 from pathlib import Path
+import unittest
```
dmehala, line 3:
```suggestion
```

test/cases/orchestration.py
```diff
@@ -788,17 +803,22 @@ def reload_nginx(self, wait_for_workers_to_terminate=True):
...
-            before = time.monotonic()
-            while old_worker_pids & nginx_worker_pids(nginx_container,
-                                                      self.verbose):
-                now = time.monotonic()
-                if now - before >= timeout_seconds:
-                    raise Exception(
-                        f"{timeout_seconds} seconds timeout exceeded while waiting for nginx workers to stop.  {now - before} seconds elapsed."
-                    )
-                time.sleep(poll_period_seconds)
+            def old_worker_stops(worker_pid):
```
pablomartinezbernardo, line 806: `old_worker_pids` is a set, and then treated as `_worker_pid` the code is doing `set not in another_set` which is always true, or am I getting something wrong?

The old code was looping while the intersection of the old pids and the current pids was non-empty, continuing once empty, which seems to be correct when expecting pids to stop, but was not really checking for new processes to spawn 🤔

dmehala: Interesting! I know David for writing high quality code, I knew I missed something, thank you for taking the time to investigate.

> the code is doing set not in another_set which is always true, or am I getting something wrong?

I intended `element not in another set` but as you mentioned `old_worker_pids` is a set. It's working because `wait_until` stop waiting when the predicate is `True` which is always the case here. 😅

dmehala: addressed in 1ce8ad96a6d298e780a14e627c0fd4630ace2a06

PR #199, opened 2025-04-16
chore(build): add nginx 1.27.5

dmehala:


.circleci/config.yml
```diff
@@ -573,14 +573,12 @@ workflows:
                             - 1.27.2
                             - 1.27.3
                             - 1.27.4
+                            - 1.27.5
                         waf:
                             - "ON"
                             - "OFF"
                 name: build << matrix.nginx-version >> on << matrix.arch >> WAF << matrix.waf >>
             - build_openresty:
-                filters:
```
dubloom, line 581: Why did you remove that ? I think it is because the generate_jobs_yaml does not have these

.circleci/config.yml
```diff
@@ -916,9 +934,6 @@ workflows:
                 requires:
                     - build << matrix.nginx-version >> on << matrix.arch >> WAF << matrix.waf >>
             - test-openresty:
-                filters:
-                    tags:
```
dubloom, line 920: Same as above

.circleci/config.yml
```diff
@@ -935,9 +950,6 @@ workflows:
                 requires:
                     - build openresty << matrix.resty-version >> on << matrix.arch >> WAF << matrix.waf >>
             - test-openresty:
-                filters:
```
dubloom, line 938: Same as above

.circleci/config.yml
```diff
@@ -954,9 +966,6 @@ workflows:
...
```
dubloom, line 957: Same as above

dubloom (commented):
I think the `generate_jobs_yaml.rb` should be updated to include the filters on openresty. Otherwise thank you for this update. 
I'll try to do the script I was working some times ago to automatically create this kind of PR when I have time

.circleci/config.yml
dmehala: addressed in 440a0d312c7f2f069e7cf4fc5b705abc0e01ce5b

dubloom (approved):
The tag is at more place than before. If this is expected, LGTM

PR #200, opened 2025-04-22
Fix WAF stalling websocket requests

cataphract:
Our body handler suppresses the flush of the upgrade response header, and the calling code ignores NGX_AGAIN (so it's pointless to change the body handler return code to it in that case).

When we finish the WAF run, we call call the write handler on the request, but the upstream connection write handler doesn't do anything because there is no data from upstream waiting to be sent.

And since we don't send the response header,  the client never sends any data, so the request stalls.

So, if we have an upstream connection (this could be generalized), replace the write handler before triggering it. This substitute write handler will attempt to flush the headers and the body filter chain before restoring the write handler.

### DIFF
--- a/src/datadog_context.h
+++ b/src/datadog_context.h
@@ -56,6 +56,10 @@ class DatadogContext {
 
   RequestTracing& single_trace();
 
+#ifdef WITH_WAF
+  security::Context* get_security_context() { return sec_ctx_.get(); }
+#endif
+
  private:
   std::vector<RequestTracing> traces_;
 #ifdef WITH_WAF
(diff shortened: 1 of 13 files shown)


src/datadog_context.h
```diff
@@ -56,6 +56,10 @@ class DatadogContext {
 
   RequestTracing& single_trace();
 
+#ifdef WITH_WAF
+  security::Context* get_security_context() { return sec_ctx_.get(); }
```
dmehala, line 60:
```suggestion
  security::Context& get_security_context() { return sec_ctx_; }
```

dmehala (approved):
LGTM! :shipit:

PR #201, opened 2025-04-22
fix(release): improve flavor handling

dmehala:
nginx-datadog#191 introduced an unwanted change for native NGINX module naming convention. This fix the regression introduced.

### DIFF
--- a/bin/release.py
+++ b/bin/release.py
@@ -171,7 +171,7 @@ def sign_package(package_path: str) -> None:
 
 def prepare_release_artifact(work_dir, build_job_number, flavor, version, arch,
                              waf):
-    flavor_prefix = f"{flavor}-"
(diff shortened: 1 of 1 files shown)


bin/release.py
```diff
@@ -298,16 +298,13 @@ def release_module(args) -> int:
...
 
-            if len(match.groups()) == 4:
-                # It is an `openresty` build
-                flavor, nginx_version, arch, waf = match.groups()
+            flavor, nginx_version, arch, waf = match.groups()
+            if flavor:
                 flavor = flavor.strip()
-            else:
-                nginx_version, arch, waf = match.groups()
-                flavor = "nginx"
```
dubloom, line 310: The default flavor name is now "" and not "nginx" anymore ? Is it the desire outcome ?

PR #204, opened 2025-04-25
RUM telemetry

pablomartinezbernardo:
Add RUM telemetry:
* All already existing metrics
* Including CSP
* Application id and remote config tags

### DIFF
--- a/src/rum/telemetry.h
+++ b/src/rum/telemetry.h
@@ -1,0 +1,35 @@
+#include <datadog/telemetry/metrics.h>
+
+#include <format>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include "version.h"
+
+namespace datadog {
+namespace nginx {
+namespace rum {
+namespace telemetry {
+
+template <typename... T>
+auto build_tags(T&&... specific_tags) {
+  std::vector<std::string> tags{std::forward<T>(specific_tags)...};
+  tags.emplace_back("integration_name:nginx");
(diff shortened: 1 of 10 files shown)


src/rum/telemetry.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include "version.h"
+
+namespace datadog {
+namespace nginx {
+namespace rum {
+namespace telemetry {
+
+template <typename... T>
+auto build_telemetry_tags(T&&... specific_tags) {
```
dmehala, line 16: I noticed that the term 'telemetry' appears both in the name and the namespace, which might be redundant. Would you consider renaming it to `build_tags` to improve clarity?

dmehala (commented):
Add few comments that should be addressed. This PR has a dependency on `dd-trace-cpp`. Please note that it cannot be approved/merged until `dd-trace-cpp` is upgraded.

pablomartinezbernardo:
All comments addressed

> Add few comments that should be addressed. This PR has a dependency on dd-trace-cpp. Please note that it cannot be approved/merged until dd-trace-cpp is upgraded.

Does this mean you are expecting dd-trace-cpp to be upgraded in this same PR? Or that given there's ongoing work in dd-trace-cpp then this has to wait?

Given RUM is not being built by default, I was leaving upgrading dd-trace-cpp up to you, and upgrading it only in the RUM branch testing it for RUM.

dmehala (approved):

PR #213, opened 2025-05-24
fix: bypass context propagation if headers are not initialized

### LINKED ISSUE #212: [BUG]: SEGFAULT when request has no headers
### Module Version(s)

1.6.1

### Bug Report

A segfault is produced In the following lines:
https://github.com/DataDog/nginx-datadog/blob/master/src/ngx_header_writer.h#L32-L34
When a request comes in with this shape
Image

It appears as if this is a GET from a client with no headers sent at all.

This causes a SEGFAULT followed by a huge increase in `nginx.net.waiting`

Image

The recoveries are after restarting nginx

### Reproduction Code

TBD

### Error Logs

```
 May 22 13:58:18 sd-prod-lb13 kernel: nginx[1476082]: segfault at 8 ip 00005575c48b2760 sp 00007ffd228d7b30 error 4 in nginx[5575c48a7000+10a000]
```

### Operating System

Ubuntu 22.04

dmehala:
When processing HTTP/0.9 requests, the HTTP headers data structure is not initialized because this version of HTTP does not support headers. This leads to a crash when the code attempts to add context propagation information to an uninitialized list of headers.

We have implemented a check to bypass context propagation if the headers are not initialized.

Resolves #212

Changes:
  - move headers logic into `common` folder.
  - add integration test with HTTP/0.9.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,7 @@ target_sources(ngx_http_datadog_objs
   PRIVATE
     src/common/variable.cpp
     src/common/directives.cpp
+    src/common/headers.cpp
     src/array_util.cpp
     src/datadog_conf.cpp
     src/datadog_conf_handler.cpp
--- a/src/common/headers.cpp
+++ b/src/common/headers.cpp
@@ -1,0 +1,68 @@
+#include "headers.h"
+
+#include "string_util.h"
+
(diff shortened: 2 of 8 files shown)


pablomartinezbernardo (approved):
LGTM

PR #216, opened 2025-06-03
Support DD_APM_TRACING_ENABLED=false

cataphract:
See https://github.com/DataDog/dd-trace-cpp/pull/217

The header injection had to be moved later, to the precontent phase, so it happens after the initial runs of the waf.


dmehala:
Implemented in #230

PR #219, opened 2025-06-10
fix(rum): glibc 2.28 compatibility

pablomartinezbernardo:
Running with RUM enabled and glibc 2.28 shows

`undefined symbol: posix_spawn_file_actions_addchdir_np`

* Enforce rust 1.74
* Add missing `fstatat` symbol

TODO
- [ ] Build and publish image once changes approved

### DIFF
--- a/build_env/Dockerfile
+++ b/build_env/Dockerfile
@@ -73,5 +73,8 @@ RUN apk add --no-cache pcre-dev pcre2-dev zlib-dev openssl-dev perl
 
 # Install Rust toolchain
 RUN apk add --no-cache curl
-RUN curl --proto '=https' –tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -yq --default-toolchain 1.74.0 \
+RUN curl --proto '=https' –tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -yq --default-toolchain 1.73.0 \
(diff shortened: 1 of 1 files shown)


pablomartinezbernardo:
The issue with using `1.73` is that `corrosion_experimental_cbindgen` always uses the latest cbindgen version (`0.29.0`), and `0.29.0` doesn't support `1.73`

> package `cbindgen v0.29.0` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0

Setting a specific cbindgen version is [not supported](https://corrosion-rs.github.io/corrosion/ffi_bindings.html#common-arguments).

What are the concerns for using `1.74`? Otherwise the only option I can think is stop using `corrosion_experimental_cbindgen` which may get messier than just using `1.74`, but very open to suggestions.

dmehala:
> corrosion_experimental_cbindgen always uses the latest cbindgen version (0.29.0)

Just to clarify, Corrosion will install the latest version of cbindgen **IF** it has not been installed yet.

> What are the concerns for using 1.74?

I don't have any particular concerns with Rust 1.74. My main point is that we were able to generate a module that loaded without any issues, so in theory, the solution *could* be as simple as using the same Rust version.

dmehala (approved):
Please enforce the same Rust version than the one required in `inject-browser-sdk` project.

PR #222, opened 2025-06-12
fix(rum): injector_end usage

pablomartinezbernardo:


xlamorlette-datadog:
This PR was created on June 22, 2025 (15 months ago) and then never updated. I’m therefore closing it as obsolete.

PR #225, opened 2025-06-24
fix(tests): bump nginx version for amazon linux

dubloom:
amazonlinux:2.0.20230418 tests are failing on master because the amazon linux nginx version was bumped to 1.28.0.
 
 This PR bumps the nginx version of the test to 1.28.0.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -475,7 +475,7 @@ workflows:
                         base-image:
                             - amazonlinux:2.0.20230418.0
                         nginx-version:
-                            - 1.26.3
+                            - 1.28.0
                         waf:
                             - "ON"
                             - "OFF"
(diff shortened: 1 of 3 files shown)


zacharycmontoya:
I've re-run the failed test runs as they seemed to fail on shutting down nginx, which seems like flake. Let's see how it goes

zacharycmontoya:
To my dismay, it seems like only re-running the failed jobs leads to this error:

```
run.py: error: argument --module-path: file /tmp/workspace/.musl-build/ngx_http_datadog_module.so does not exist
```

It seems like a full re-run is needed, so I'll do that now.

zacharycmontoya:
After a re-run, we're still receiving issues about a timeout:

```
Exception: 10 seconds timeout exceeded while waiting for nginx workers to stop.  10.086069895999998 seconds elapsed.
```

Since this appears to be flake, I'll ignore it for now.

zacharycmontoya (approved):

dmehala:
Closing because we are not supporting this version of AL anymore. Thanks for you contribution @dubloom .

PR #220, opened 2025-06-10
chore(benchmarks): add gate and notify jobs in macrobenchmarks

dubloom:

PR #226, opened 2025-06-24
chore(benchmarks): run macrobenchmark on last changes

dubloom:
Before this PR, the macrobenchmarks were run only on the latest release which is highly reducing their utility.

This PR runs the macrobenchmarks on the latest changes. 

**Important**: macrobenchmarks will fail until this PR is merged because changes were made also in the benchmarking-platform repo.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -23,7 +23,7 @@ build-nginx-module:
     COVERAGE: OFF
     RUM: OFF
     BUILD_TYPE: Release
-    ARCH: amd64
+    ARCH: x86_64
   script:
     - git submodule sync && git submodule update --init --recursive
     - export NGINX_SRC_DIR="$PWD/nginx"
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -1,4 +1,5 @@
 stages:
+  - build
   - run-benchmarks
   - gate
   - notify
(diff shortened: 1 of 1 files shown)


.gitlab/benchmarks.yml
```diff
@@ -9,13 +10,40 @@ include:
...
+
 
 .benchmarks:
   stage: run-benchmarks
+  needs: ["build-nginx-module"]
   tags: ["runner:apm-k8s-same-cpu"]
   timeout: 1h
   image: $MACROBENCHMARKS_CI_IMAGE
   script:
+    - cp .musl-build/ngx_http_datadog_module.so /usr/lib/nginx/modules/ngx_http_datadog_module.so
```
zacharycmontoya, line 46: I understand that this stage now triggers a build and uses that artifact to be tested, but what I don't understand is where the change is made that will trigger a GitLab pipeline to run on every commit instead of every release, which I don't even know if it's automatically done or manually done. Can you help me understand that?

ddyurchenko (approved):

igoragoli (commented):
Hey @dubloom, just a quick thing that went under my radar on https://github.com/DataDog/nginx-datadog/pull/220 and that we could potentially include in this PR.

The `check-slo-breaches` image should be set to `registry.ddbuild.io/images/benchmarking-platform-tools-ubuntu:latest` so that the threshold comparison tooling is up to date.

I'm setting up Slack notifications for cases other than breaches/warnings so we know early when a gating job is failing for other reasons (like not running benchmarks on latest changes).

.gitlab/benchmarks.yml
```diff
@@ -8,14 +9,41 @@ include:
  file: 'images/templates/gitlab/notify-slo-breaches.template.yml'
 
 variables:
-  MACROBENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:cpp-nginx
+  MACROBENCHMARKS_CI_IMAGE: registry.ddbuild.io/images/benchmarking-platform-tools-ubuntu:latest
```
ddyurchenko, line 12: This was correct, the image here should remain `MACROBENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:cpp-nginx`

dmehala: image

Ok, I have no idea what's going on at this point

dmehala:
```suggestion
  MACROBENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:cpp-nginx
```

dmehala: fixed in 0b672c6ca6c39d402503b490a39cf214d94a04e8

.gitlab/benchmarks.yml
zacharycmontoya: Ah I missed the fact that this is included in `~/.gitlab-ci.yml` and triggered when `$CI_COMMIT_REF_NAME == "master"`. Makes sense now 👍🏼

PR #228, opened 2025-07-08
ci: support latest releases of nginx and ingress-nginx

dmehala:
This commit ensures we build the module for the latest version of NGINX and ingress-nginx.

Changes:
  - Deprecates NGINX versions that fall outside our compatibility window.
  - Introduces support for NGINX 1.19.0.
  - Add support for ingress-nginx 1.11.8, 1.12.4 and 1.13.0


.circleci/src/workflows/build-and-test.yml
```diff
@@ -19,11 +19,10 @@ jobs:
         - amd64
         - arm64
         nginx-version:
-        - 1.26.3
         - 1.24.0
         - 1.25.4
-        - 1.27.5
         - 1.28.0
+        - 1.29.0
```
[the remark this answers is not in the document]
dmehala: `build-and-test` is the workflow ran on `master` and PRs. We mainly run on mainline and the latest stable version of NGINX.

.circleci/src/workflows/build-and-test-all.yml
```diff
@@ -134,13 +135,31 @@
             - 'OFF'
             base-image:
             - amazonlinux:2023.3.20240219.0
```
zacharycmontoya, line 137: For my own education, is there a set of base images that we aim to support? And if Amazon Linux is one of them, is that the only reason we are still testing NGINX 1.24, which would otherwise be out of support?

dmehala: > Amazon Linux is one of them, is that the only reason we are still testing NGINX 1.24, which would otherwise be out of support?
yes.

zacharycmontoya (approved):

PR #232, opened 2025-07-21
waf: implement response body collection

cataphract:
Submit response body to the WAF (if it's available and parseable). In this case, the final WAF run is delayed until enough data is gotten.

Also do some light refactoring: in security/context.cpp, move some ddwaf functionality to a new file.

### DIFF
--- a/src/security/context.cpp
+++ b/src/security/context.cpp
@@ -1,10 +1,9 @@
 #include "context.h"
 
 #include <atomic>
-#include <charconv>
+#include <climits>
 #include <optional>
 #include <stdexcept>
-#include <string>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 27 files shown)


src/security/context.cpp
```diff
@@ -332,6 +226,14 @@ class PolTaskCtx {
     ngx_log_debug(NGX_LOG_DEBUG_HTTP, req_log(), 0,
                   "refcount before decrement upon task %p completion: %d", this,
                   count);
+
+    // create a stack copy to safely call destructor later
+    // the task is allocated on the request pool, so by the end of this method
+    // maybe *this is no longer valid
+    alignas(Self) char self_copy_storage[sizeof(Self)];
+    std::memcpy(self_copy_storage, this, sizeof(Self));
+    Self *self_copy = reinterpret_cast<Self *>(self_copy_storage);
```
Anilm3, line 235: Are you suggesting that on request pool destruction, the destructor of this class isn't called?

This is quite sketchy, wouldn't it be better to have the request pool hold a pointer so you can just use `this` to invoke the destructor?

src/security/context.cpp
```diff
@@ -873,12 +603,42 @@ class PolFinalWafCtx : public PolTaskCtx<PolFinalWafCtx> {
...
+    // flush the buffered header.
+    // (If the submission fails, this is not strictly required, because we could
+    // flush the buffered header synchronously, but do it all the same to avoid
+    // multiplying code paths)
+    if (from_header_filter && req.header_only) {
+      ngx_log_debug(NGX_LOG_DEBUG_HTTP, req.connection->log, 0,
+                    "suppressing req.header_only");
+      req.header_only = 0;
+    }
+    req.header_sent = 1;  // skip/alert when attempting to sent headers
```
Anilm3, line 623:
```suggestion
    req.header_sent = 1;  // skip/alert when attempting to send headers
```

src/security/context.cpp
```diff
@@ -962,7 +722,7 @@ ngx_int_t Context::do_request_body_filter(ngx_http_request_t &request,
       auto *conf = static_cast<datadog_loc_conf_t *>(
           ngx_http_get_module_loc_conf(&request, ngx_http_datadog_module));
 
-      if (task_ctx.submit(conf->waf_pool)) {
+      if (std::move(task_ctx).submit(conf->waf_pool)) {
```
Anilm3, line 725: I assume this is related to what I comment on above but perhaps it deserves a comment / explanation here.

src/security/context.cpp
```diff
@@ -1500,26 +1260,26 @@ ngx_int_t Context::do_header_filter(ngx_http_request_t &request,
...
-  transition_to_stage(stage::PENDING_WAF_END);
-
   ngx_log_debug(NGX_LOG_DEBUG_HTTP, request.connection->log, 0,
-                "waf header filter: waf end task; replacing send_chain handler "
+                "waf header filter: replacing send_chain handler "
                 "and invoking the next header filter");
 
+  if (request.header_only) {
+    header_only_ =
+        true;  // save it; we'll change req.header_only upon waf task submission
```
Anilm3, line 1279: nit: perhaps put the comment above

src/security/context.cpp
```diff
@@ -1537,25 +1297,41 @@ ngx_int_t Context::do_header_filter(ngx_http_request_t &request,
...
+  // were unable to submit the WAF task)
+  waf_send_resp_body_ = waf_send_resp_body_ && is_body_resp_parseable(request);
+  if (!waf_send_resp_body_) {
     ngx_log_debug(NGX_LOG_DEBUG_HTTP, request.connection->log, 0,
-                  "waf header filter: ngx_http_write_filter returned %d", rc);
-    return rc;
+                  "waf header filter: downstream filters returned NGX_OK; "
+                  "attempting to submit WAF task (waf_send_resp_body=%d, "
+                  "header_only=%d, content_length=%d)",
+                  waf_send_resp_body_, request.header_only,
```
Anilm3, line 1311: nit: you already know beforehand the value of `waf_send_resp_body_`

src/security/context.cpp
```diff
@@ -1776,23 +1577,28 @@ void Context::do_on_main_log_request(ngx_http_request_t &request,
...
-  if (results_.empty()) {
-    return;
-  }
+  auto &&trace_segment = span.trace_segment();
 
-  report_match(request, span.trace_segment(), span, results_);
-  results_.clear();
+  bool did_report = waf_ctx_->report_matches([&](std::string_view json) {
+    ngx_str_t json_ns{dnsec::ngx_stringv(json)};
+    ngx_log_error(NGX_LOG_INFO, request.connection->log, 0, "appsec event: %V",
```
Anilm3, line 1590: Events should be logged in debug / trace only.

Anilm3 (approved):
It's a bit hard to follow, but it looks generally correct.

PR #236, opened 2025-08-04
Fix API Doc: swap datadog_environment and datadog_version descriptions back

gaffneyd4:
The description for datadog_environment was under datadog_version and the description for datadog_version was under datadog_environment. This PR swaps them back so they match the correct directives.


dmehala (approved):
Good catch! Thank you for your contribution.

PR #242, opened 2025-08-18
Implement libddwaf schema collection

cataphract:
Also update libddwaf.

### DIFF
--- a/src/security/context.cpp
+++ b/src/security/context.cpp
@@ -1591,8 +1591,8 @@ std::optional<BlockSpecification> Context::run_waf_end(
   return block_spec;
 }
 
-bool Context::has_matches() const noexcept {
-  return waf_ctx_ && waf_ctx_->has_matches();
+bool Context::keep_span() const noexcept {
+  return waf_ctx_ && waf_ctx_->keep();
 }
 
 void Context::on_main_log_request(ngx_http_request_t &request,
--- a/src/security/ddwaf_req.cpp
+++ b/src/security/ddwaf_req.cpp
@@ -397,13 +397,24 @@ DdwafContext::WafRunResult DdwafContext::run(ngx_log_t &log,
   return waf_result;
 }
 
+bool DdwafContext::has_matches() const {
+  for (const libddwaf_owned_ddwaf_obj<ddwaf_map_obj> &result : results_) {
+    std::optional<ddwaf_arr_obj> maybe_events =
+        result.get_opt<ddwaf_arr_obj>("events");
+    if (maybe_events && maybe_events->size() > 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
 bool DdwafContext::report_matches(
     const std::function<void(std::string_view)> &f) {
   if (results_.empty()) {
     return false;
   }
 
-  std::vector<ddwaf_arr_obj *> events_arrs;
+  std::vector<ddwaf_arr_obj> events_arrs;
   for (LibddwafOwnedMap &result : results_) {
     std::optional<ddwaf_arr_obj> maybe_events =
         result.get_opt<ddwaf_arr_obj>("events");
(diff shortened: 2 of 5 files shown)


src/ngx_http_datadog_module.cpp
```diff
@@ -330,6 +333,16 @@ static ngx_int_t datadog_module_init(ngx_conf_t *cf) noexcept {
     }
   }
 
+#ifdef WITH_WAF
+  // Initialize shared memory for API security rate limiter
+  if (security::Library::initialize_api_security_shared_memory(cf) != NGX_OK) {
+    ngx_log_error(NGX_LOG_WARN, cf->log, 0,
+                  "Failed to initialize API security shared memory, will use "
+                  "per-process limiter");
+    return NGX_ERROR;
```
Anilm3, line 342: Does it make sense to return an error if you have a fallback?

src/security/compress.h
```diff
@@ -0,0 +1,50 @@
...
+
+extern "C" {
+#define ZLIB_CONST
+#include <zlib.h>
+}
+
+namespace {
+size_t estimate_compressed_size(size_t in_len) {
+  // NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+  return (((size_t)((double)in_len * (double)1.015)) + 10 + 8 + 4 + 1);
```
Anilm3, line 12: nit: perhaps a comment on where this comes from could be useful.

```
// Taken from PHP approach
// https://heap.space/xref/PHP-7.3/ext/zlib/php_zlib.h?r=8d3f8ca1#36

File: src/security/context.cpp [hunk 3]
```diff
@@ -1564,10 +1581,19 @@ std::optional<BlockSpecification> Context::run_waf_end(
...
+  ddwaf_object *resp_data = collect_response_data(
+      request, body_chain, body_size, extract_schema, memres_);
+
+  auto &&log = *request.connection->log;
+  auto [_, block_spec, tags] = waf_ctx_->run(log, *resp_data);
+
+  for (auto &[key, json] : tags) {
+    ngx_log_debug(NGX_LOG_DEBUG_HTTP, &log, 0, "Setting tag %s with value %s",
+                  key.c_str(), json.c_str());
+    span.set_tag(key, json);
```
Anilm3, line 1594: Should be checking the tag prefix (it could be a metric).

src/security/ddwaf_req.cpp
```diff
@@ -238,6 +248,60 @@ std::optional<BlockSpecification> resolve_block_spec(
...
 
+namespace {
+constexpr int kMaxPlainSchemaAllowed = 260;
+constexpr int kMaxSchemaSize = 25000;
+
+template <typename Func>
+void handle_attribute(ngx_log_t &log, const ddwaf_obj &obj, Func &&f) {
+  ngx_str_t key = to_ngx_str(obj.key());
+
+  if (!obj.key().starts_with("_dd.appsec.s."sv)) {
```
Anilm3, line 259: Why do you ignore other attributes, I understand you don't want to compress / encode them but at least they should be serialised as tags / metrics.

src/ngx_http_datadog_module.cpp
Anilm3: I didn't see a per-process limiter so perhaps the message is outdated.

src/security/shared_limiter.h
```diff
@@ -0,0 +1,226 @@
...
+  std::uint32_t tokens;
+  std::uint32_t last_refresh;
+};
+
+static_assert(std::atomic<TokensAndRefresh>::is_always_lock_free,
+              "TokensAndRefresh must be lock-free");
+
+template <std::uint32_t RefreshesPerMin>
+struct SharedLimiterState {
+  std::atomic<TokensAndRefresh> tokens_and_refresh{};
```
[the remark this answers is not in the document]
cataphract: There's a [static assertion](https://github.com/DataDog/nginx-datadog/pull/242/files#diff-a92d8d5be3799e8a246e776747293ecc319b3a0064cc269997e876db026ed36aR24-R25) to that effect.

PR #245, opened 2025-08-29
fix: Correctly handle auth requests when subrequest logging is enabled

zacharycmontoya:
Fixes a regression introduced by https://github.com/DataDog/nginx-datadog/pull/231

## Scenario
When NGINX is configured in a way to send subrequests, such as [Authorization Based on Subrequest Result](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-subrequest-authentication/), we generally do not create a span for it. However, when NGINX is configured with `log_subrequest on;`, we create a `RequestTracing` object (and a span) for it, which causes a crashing error when we run the new assertion that there may only be [one](https://github.com/DataDog/nginx-datadog/blob/1aadb4be81dd5935a90aedfdf5c7f96c1268e871/src/datadog_context.cpp#L255) `RequestTracing` object present.

## Fix
Change the behavior so that when we are deciding what span context to inject into headers and register with the security context we 1) do not assert there's only one span and 2) we use the context from the root span.

## Additional notes
In the PR I introduced a regression test to match the previous v1.6.2 nginx-datadog behavior. However, that showed that downstream span would become a child span of the auth subrequest instead of the root nginx span. This PR makes a breaking change so that:
1. The downstream span is a child of the root nginx span (instead of the auth subrequest span)
2. The security context is added to the root span (instead of the auth subrequest span)

### DIFF
--- a/src/datadog_context.cpp
+++ b/src/datadog_context.cpp
@@ -343,7 +343,7 @@ ngx_int_t DatadogContext::on_precontent_phase(ngx_http_request_t *request) {
   // inject headers in the precontent phase into the request headers
   // These headers will be copied by ngx_http_proxy_create_request on the
   // content phase into the outgoing request headers (probably)
-  RequestTracing &trace = traces_.back();
+  RequestTracing &trace = traces_.front();
   dd::Span &span = trace.active_span();
   span.set_tag("span.kind", "client");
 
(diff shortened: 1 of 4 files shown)


zacharycmontoya:
> I think it's fine as a fix, although I wonder why we create more than one trace, instead of making the new request just a new span under the current trace

We do exactly as you say, we have one trace with two spans. However, I think there's some follow-up we can do to improve how we are handling the two separate spans and their data structures to improve our handling of these subrequest.

PR #255, opened 2025-09-23
fix: script returning NGINX tarball source code

### LINKED ISSUE #243: [BUG]: The curl to get the nginx version is always empty
### Module Version(s)

1.6.2

### Bug Report

# Bug: nginx_release_downloads.sh returns empty string for version argument

## Description
When running `bin/nginx_release_downloads.sh 1.28.0`, the script returns an empty string instead of the expected download URL. This causes subsequent curl commands to fail with the error:
curl: option : blank argument where content is expected

## Steps to Reproduce
1. Run the command: `bin/nginx_release_downloads.sh 1.28.0`
2. Observe that it returns an empty string instead of a valid URL
3. The failed output is then used in: `curl -s -S -L -o nginx.tar.gz ""`

## Affected Versions
- Version 1.6.2 ✅
- Version 1.7.0 ✅
- Master branch ❌ (not affected - the issue has been fixed but introduces other breaking changes)

## Root Cause
The script `bin/nginx_release_downloads.sh` does not parse the version argument correctly. The filter function has issues with:
- Incorrect regex pattern matching
- Improper handling of version argument boundaries
- Use of `tail -1` instead of `head -1` for version selection

## Expected Behavior
The script should return a valid download URL when provided with a version argument, for example:
https://nginx.org/download/nginx-1.28.0.tar.gz

## Additional Context
I tested with master branch it resolves this issue but introduces breaking changes in other parts of the codebase.

Let me know if you need any additional information.

##Script Used
this is our script to build the modules

```
#!/bin/bash

....

git clone --branch v1.6.2 https://github.com/DataDog/nginx-datadog.git
cd nginx-datadog/ || exit 1

NGINX_VERSION=$(get_nginx_version)

echo "Nginx version detected: ${NGINX_VERSION}"

NGINX_VERSION=${NGINX_VERSION} make build || exit 2

mv .build/ngx_http_datadog_module.so /etc/nginx/modules/ngx_http_datadog_module.so || exit 3

cd ../

rm -rf nginx-datadog/ || exit 4

.....
```

### Reproduction Code

_No response_

### Error Logs

_No response_

### Operating System

_No response_

dmehala:
Resolves #243


dubloom (commented):
I'm not sure to follow the changes here. 

Removing the script will not resolve the issue, it will just prevent it from happening. However, the issue reporter will still not be able to achieve what he wants.

dubloom (approved):
LGTM, was the script described somewhere ? (to know if we should update the doc)

PR #262, opened 2025-10-23
Adds a weekly scheduled job to the CI

cbeauchesne:
system-tests requires an artifact to execute test. CircleCI gots an artifact's retention period of 9 days, so if there is no activity on the repo for 9 days, one needs to manually re-run the job.

This PR automate this by adding a weekly scheduled job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -367,13 +367,6 @@ parameters:
 version: 2.1
 workflows:
     build-and-test:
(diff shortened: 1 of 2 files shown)


dmehala (commented):
`config.yml` is a generated file. Please update the workflow in `.circleci/workflow` and run `make circleci-config` to update the generated file.

PR #274, opened 2025-12-16
chore(ci): downsize test executors

dmehala:


brettlangdon (approved):
nice, thanks!

I'll be interested to see the resource utilization after this change, if we can go even lower, or if we went too low.

PR #350, opened 2026-04-15
Add support for Nginx 1.30.0

xlamorlette-datadog:
- Add support for [Nginx 1.30.0](https://github.com/nginx/nginx/releases/tag/release-1.30.0).
- Bump version to `1.16.0`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 cmake_policy(SET CMP0068 NEW)
 cmake_policy(SET CMP0135 NEW)
 
-set(NGINX_DATADOG_VERSION 1.15.0)
+set(NGINX_DATADOG_VERSION 1.16.0)
 project(ngx_http_datadog_module VERSION ${NGINX_DATADOG_VERSION})
 
 option(NGINX_DATADOG_ASM_ENABLED "Build with libddwaf" ON)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 cmake_policy(SET CMP0068 NEW)
 cmake_policy(SET CMP0135 NEW)
 
-set(NGINX_DATADOG_VERSION 1.15.0)
+set(NGINX_DATADOG_VERSION 1.16.0)
```
dubloom, line 10: Why are we bumping this ?

xlamorlette-datadog: Good question!

To add the support of a new version of Nginx, we have to build a new release.

Adding the support of a new version of Nginx based on an already delivered version of Nginx Datadog would be more complex:
- We should create a branch for the latest release (even if we think no functional change has been made since the latest release, you can't be 100% certain).
- We would have to modify our delivery script to deliver the binaries only for the added Nginx version.
- We should add a way to track the support of the Nginx versions in the different Nginx Datadog releases (currently, this is simply in the release notes).
- Strictly speaking, for a given Nginx Datadog versions, the binaries for the different Nginx versions would be based on different commits, which may lead to issues.

I agree that this leads to bump the version with no functional change, which can be a bit annoying for users using an old Nginx version, and so who don't need to upgrade to the latest Nginx Datadog release.

dubloom: I was mostly checking if this was a leftover. It seems there is a reason why we should bump it so I trust you.

dubloom (approved):

CMakeLists.txt
xlamorlette-datadog: This is really a good question, thanks for raising it! I'm copying this at the bottom of the page describing the delivery procedure.

PR #366, opened 2026-05-06
Release v1.16.1

xlamorlette-datadog:
Prepare release `v1.16.1`, with the following changes:

- Update `dd-trace-cpp` to [v2.1.0](https://github.com/DataDog/dd-trace-cpp/releases/tag/v2.1.0).
- Remove support of Nginx 1.24.0 (which reached end of life more than one year ago, on May 1st 2025).
- Upgrade Amazon Linux to latest available version (we used a 2 years old that embedded Nginx `1.24.0`).

### DIFF
--- a/.gitlab/build-and-test-all.yml
+++ b/.gitlab/build-and-test-all.yml
@@ -302,8 +302,8 @@ test-nginx-all-bis:
         NGINX_VERSION: ["1.30.0"]
         WAF: ["ON", "OFF"]
       - ARCH: ["amd64", "arm64"]
-        BASE_IMAGE: ["amazonlinux:2023.3.20240219.0"]
-        NGINX_VERSION: ["1.30.0"]
+        BASE_IMAGE: ["amazonlinux:2023.11.20260427.1"]
+        NGINX_VERSION: ["1.28.3"]
         WAF: ["ON", "OFF"]
 
 test-nginx-rum-all:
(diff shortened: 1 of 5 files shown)


zacharycmontoya:
> _Under investigation_: Upgrading Nginx from `1.24.0` to `1.30.0` for the tests with `amazonlinux` fails…

@xlamorlette-datadog if this is leading to failures, we can also consider separating this into a separate PR

xlamorlette-datadog:
@zacharycmontoya 
> if this is leading to failures, we can also consider separating this into a separate PR

This comes with the removal of the support of Nginx 1.24.0. I think I could get a fix within a reasonable timeframe by upgrading the used `amazonlinux` image (by the way, is this test really needed?…). Else, I will revert this change…

pawelchcki (approved):

zacharycmontoya:
> @zacharycmontoya
> 
> > if this is leading to failures, we can also consider separating this into a separate PR
> 
> This comes with the removal of the support of Nginx 1.24.0. I think I could get a fix within a reasonable timeframe by upgrading the used `amazonlinux` image (by the way, is this test really needed?…). Else, I will revert this change…

@xlamorlette-datadog yep I understand that this came with the removal of Nginx 1.24.0. If it takes more than an hour to resolve, I suggest reverting the change and we can tackle separately

PR #391, opened 2026-05-27
fix: check buf->last_buf instead of buf->last in chain::has_last()

cataphract:
Replaces https://github.com/DataDog/nginx-datadog/pull/386 to add a commit, satisfy the signing requirements and run the CI jobs.

Key elements from the original PR:

## Impact
Any upstream response with `Content-Length: 0` and a parseable `Content-Type` (`application/json`, `text/plain`) hangs for 60 seconds and returns `408` when AppSec is enabled.

## Reproduction

1. Configure Nginx with AppSec enabled (`datadog_appsec_enabled on`).
2. Set up a `proxy_pass` upstream that returns `202 Accepted` with `Content-Length: 0` and `Content-Type: application/json`.
3. Send a POST request through Nginx.
4. Observe: Nginx hangs for 60s (`send_timeout` default), then returns 408 with `bytes_sent: 0`.

### DIFF
--- a/src/security/util.h
+++ b/src/security/util.h
@@ -211,7 +211,7 @@ inline std::size_t has_special(ngx_chain_t const *ch) {
 }
 inline std::size_t has_last(ngx_chain_t const *ch) {
   for (ngx_chain_t const *cl = ch; cl; cl = cl->next) {
-    if (cl->buf->last) {
+    if (cl->buf->last_buf) {
       return true;
     }
   }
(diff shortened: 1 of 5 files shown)


xlamorlette-datadog (approved):

xlamorlette-datadog:
@moneyforward-hoai-nam-dao Your bug fix is available in https://github.com/DataDog/nginx-datadog/releases/tag/v1.19.1. Thanks again!

PR #395, opened 2026-06-03
Fix nullopt deref / add asan run with libcpp debug assertions

cataphract:
See APMSP-3401

For ASAN, we need to build nginx and the module with sanitizers and dynamically link against libc++; if there is more than one copy of the c++ runtime (e.g. pulled through libasan), we get into problems. For this, the build image was changed: now it also builds a shared lib libc++/libc++api and it builds libasan against it as well.


.gitlab/build-and-test-fast.yml
```diff
@@ -66,6 +66,16 @@ build-nginx-fast:
           - "1.31.1"
         WAF: ["ON", "OFF"]
 
+build-nginx-asan-fast:
+  extends:
+    - .build-and-test-fast
+    - .build-nginx-asan
+  variables:
+    ARCH: "amd64"
+    NGINX_VERSION: "1.28.3"
```
xlamorlette-datadog: Rather use the latest available Nginx at the time of writing: `1.31.1`.
(Same, of course, line 168 in `test-nginx-asan-fast`.)

Makefile
```diff
@@ -7,7 +7,10 @@ COVERAGE ?= OFF
 MAKE_JOB_COUNT ?= $(shell nproc)
 PWD ?= $(shell pwd)
 RUM ?= OFF
+ASAN ?= OFF
```
xlamorlette-datadog: _nit, pedantic_: please keep variables sorted alphabetically

.gitlab/common.yml
```diff
@@ -79,6 +93,17 @@ variables:
     paths:
       - test/logs/test.log
 
+.test-nginx-asan:
+  extends: .test
+  variables:
+    ASAN: "ON"
```
xlamorlette-datadog: I think the `ASAN` variable is useless in this context, because the job only calls `test --asan`.

.gitlab/common.yml
```diff
@@ -32,6 +32,20 @@ variables:
...
       - artifacts/$ARCH/$NGINX_VERSION/$WAF/
 
+.build-nginx-asan:
+  extends: .build
+  variables:
+    RUM: "OFF"
+    ASAN: "ON"
+    CMAKE_BUILD_TYPE: "Debug"
+  script:
+    - make BUILD_TYPE="$CMAKE_BUILD_TYPE" build-musl-aux
```
xlamorlette-datadog: _nit_: You could simplify by naming the GitLab variable `BUILD_TYPE`, as used in the `Makefile` (as for `RUM` and `ASAN`):
```
  BUILD_TYPE: "Debug"
  script:
    - make build-musl-aux
```

.gitlab/build-and-test-fast.yml
```diff
@@ -66,6 +66,16 @@ build-nginx-fast:
...
+    WAF: "ON"
+    CMAKE_BUILD_TYPE: "Debug"
```
xlamorlette-datadog: This is already set in `.build-nginx-asan`.

.gitlab/benchmarks.yml
```diff
@@ -13,7 +13,7 @@ include:
 
 variables:
   MACROBENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:cpp-nginx
-  BUILD_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/images/mirror/b1o7r7e0/nginx_musl_toolchain
+  BUILD_IMAGE: registry.ddbuild.io/ci/nginx-datadog/nginx_musl_toolchain@sha256:432612f8c66c51cbee7134bde4b611df04f77a9b4c2b22c691bcbec12dfcf4a9
```
xlamorlette-datadog, line 16: Are we sure we want to change the Docker image for benchmarks?
I don't remember why it used to be in `486234852809.dkr.ecr.us-east-1.amazonaws.com` rather than `registry.ddbuild.io/ci/nginx-datadog`…

build_env/Dockerfile
```diff
@@ -18,6 +18,9 @@ RUN wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${LLVM_V
     grep -F llvm-project-${LLVM_VERSION}.src.tar.xz /CHECKSUMS | sha512sum --check && \
     tar -xvf llvm-project-${LLVM_VERSION}.src.tar.xz
 
+COPY compiler-rt-musl-tls.diff /compiler-rt-musl-tls.diff
+RUN cd llvm-project-${LLVM_VERSION}.src && patch -p1 < /compiler-rt-musl-tls.diff
```
xlamorlette-datadog, line 28: May you please add a little comment / reference why we need this patch?
As far as I can understand, this fixes a thread-local storage (TLS) incompatibility between compiler-rt and musl libc.

test/bin/run.py
```diff
@@ -52,40 +54,128 @@ def validate_file(arg):
     return arg
 
 
+def format_cmd(cmd):
```
xlamorlette-datadog: _nit_: add type hints for function signatures:
```
def format_cmd(cmd: list[str]) -> str:
```

test/bin/run.py
```diff
@@ -52,40 +54,128 @@ def validate_file(arg):
...
+    return " ".join(shlex.quote(str(part)) for part in cmd)
+
+
+def normalize_arch(machine):
```
xlamorlette-datadog: _nit_: type hints:
```
def normalize_arch(machine: str) -> str:
```

test/bin/run.py
```diff
@@ -52,40 +54,128 @@ def validate_file(arg):
...
+    if machine in ("arm64", "aarch64"):
+        return "aarch64"
+    if machine in ("amd64", "x86_64"):
+        return "x86_64"
+    return machine
+
+
+def docker_platform_for_arch(arch):
```
xlamorlette-datadog: _nit_: type hints:
```
def docker_platform_for_arch(arch: str) -> str:
```

test/bin/run.py
```diff
@@ -52,40 +54,128 @@ def validate_file(arg):
...
+    if arch == "aarch64":
+        return "linux/arm64"
+    if arch == "x86_64":
+        return "linux/amd64"
+    raise SystemExit(f"--arch: unsupported architecture {arch!r}")
+
+
+def build_asan_nginx_base(arch):
```
xlamorlette-datadog: _nit_: type hints and comment:
```
def build_asan_nginx_base(arch: str) -> str:
    """Returns the Docker image name."""
```

test/bin/run.py
```diff
@@ -52,40 +54,128 @@ def validate_file(arg):
...
+        f"--build-arg NGINX_VERSION={shlex.quote(nginx_version)} "
+        f"--build-arg ARCH={shlex.quote(arch)} "
+        f"--build-arg TOOLCHAIN_IMAGE={shlex.quote(toolchain_image)} "
+        f"-f {shlex.quote(os.path.join(nginx_service_dir, 'Dockerfile.asan-base'))} "
+        f"-t {shlex.quote(tag)} "
+        f"{shlex.quote(nginx_service_dir)}")
+    return tag
+
+
+def run_unittest(inputs, timeout_seconds, cwd):
```
xlamorlette-datadog: _nit_: type hints:
def run_unittest(inputs: list[str], timeout_seconds: int, cwd: str):

test/cases/orchestration.py
```diff
@@ -155,6 +155,7 @@ def docker_compose_ps(service):
                             stdout=subprocess.PIPE,
                             env=child_env(),
                             encoding="utf8",
+                            errors="replace",
```
xlamorlette-datadog, line 158: _out of scope_: only for the record: we have many `subprocess.run()` calls with similar arguments in this file; this could be factorized.

test/cases/orchestration.py
```diff
@@ -426,7 +432,11 @@ def add_services_in_nginx_etc_hosts(services):
     # "-T" means "don't allocate a TTY".  This is necessary to avoid the
     # error "the input device is not a TTY".
     command = docker_compose_command("exec", "-T", "--", "nginx", "/bin/sh")
-    subprocess.run(command, input=script, env=child_env(), encoding="utf8")
+    subprocess.run(command,
+                   input=script,
+                   env=child_env(),
+                   encoding="utf8",
+                   errors="replace")
```
xlamorlette-datadog: _nit_: By the way, I think we should add `check=True` to raise failures, as in other `subprocess.run()` calls.

.gitlab/build-and-test-fast.yml
```diff
@@ -40,7 +40,7 @@ lint:
 
 shellcheck:
   extends: .build-and-test-fast
-  image: $CI_REGISTRY/nginx_musl_toolchain:rust-1.94
+  image: $CI_REGISTRY/nginx_musl_toolchain@sha256:432612f8c66c51cbee7134bde4b611df04f77a9b4c2b22c691bcbec12dfcf4a9
```
xlamorlette-datadog: _nit_: This image digest is repeated in several yml files. We could factorize this as a variable in `common.yml`:
```
variables:
  CI_REGISTRY: "registry.ddbuild.io/ci/nginx-datadog"
  TESTS_IMAGES_REGISTRY: "$CI_REGISTRY/mirror"
  TOOLCHAIN_IMAGE: "$CI_REGISTRY/nginx_musl_toolchain@sha256:432612f8c66c51cbee7134bde4b611df04f77a9b4c2b22c691bcbec12dfcf4a9"
```

And then use it as follows:
- `common.yml` line 101: `ASAN_TOOLCHAIN_IMAGE="$TOOLCHAIN_IMAGE"`
- `build-and-test-fast.yml` line 43: `image: $TOOLCHAIN_IMAGE`
- `benchmarks.yml` line 16: `BUILD_IMAGE: $TOOLCHAIN_IMAGE`
- `ssi-package.yml` line 13: `image: $TOOLCHAIN_IMAGE`

CMakeLists.txt
```diff
@@ -207,6 +208,16 @@ if(ENABLE_FUZZERS)
   target_compile_options(ngx_http_datadog_objs PUBLIC -fsanitize=fuzzer,address,undefined,leak)
 endif()
 
+if(ENABLE_ASAN)
+  target_compile_options(ngx_http_datadog_objs PUBLIC -fsanitize=address,undefined)
+  target_compile_definitions(ngx_http_datadog_objs PUBLIC
+    $<$<COMPILE_LANGUAGE:CXX>:_LIBCPP_ENABLE_ASSERTIONS=1>)
+  target_link_options(ngx_http_datadog_objs INTERFACE -fsanitize=address,undefined)
```
xlamorlette-datadog: To be checked: this could be redundant with the `target_link_options(ngx_http_datadog_module PRIVATE -fsanitize=address,undefined -shared-libasan)` below.

xlamorlette-datadog (approved):
Thanks a lot for this really great addition!!!

.gitlab/benchmarks.yml
cataphract: it should be able to use this image, which is better than dealing with multiple images

xlamorlette-datadog: I'm just a bit concerned that it could change the baseline for the benchmarks. I can monitor this.

PR #399, opened 2026-06-09
Add clang plugin to detect ngx_log mismatches and fix the existing ones

cataphract:


cataphract:
@codex review

bin/lint.sh
```diff
@@ -14,7 +14,7 @@ fi
 
 error_messages=''
 
-find src/ -type f \( -name '*.h' -o -name '*.cpp' -o -name '*.c' \) -print0 | xargs -0 clang-format-14 --Werror --dry-run --style=file
+find src/ tools/clang-tidy -type f \( -name '*.h' -o -name '*.cpp' -o -name '*.c' \) -print0 | xargs -0 clang-format-14 --Werror --dry-run --style=file
```
xlamorlette-datadog: _nit_: tiny simplification: `tools` instead of `tools/clang-tidy`.
(By the way, we should also add `test`. I will do this in a dedicated PR. I may also try to factorize some stuff.)

bin/format.sh
```diff
@@ -11,7 +11,7 @@ if ! [ -e .clang-format ]; then
     exit 1
 fi
 
-find src/ -type f \( -name '*.h' -o -name '*.cpp' -o -name '*.c' \) -print0 | xargs -0 clang-format-14 -i --style=file
+find src/ tools/clang-tidy -type f \( -name '*.h' -o -name '*.cpp' -o -name '*.c' \) -print0 | xargs -0 clang-format-14 -i --style=file
```
xlamorlette-datadog: _nit_: tiny simplification: `tools` instead of `tools/clang-tidy`.
Same in `bin/lint.sh` line 17.
(By the way, we should also add `test`. I will do this in a dedicated PR.)

bin/nginx-log-format-tidy.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
+set -eo pipefail
+
+container_repo=/repo
+
+if [ "${NGINX_LOG_FORMAT_TIDY_IN_CONTAINER:-}" != "1" ]; then
```
xlamorlette-datadog: I like this technique which I didn't know. Thanks, I learnt something!

_nit_: unset variable expands to empty string (unless using `set -u`) so this can be simplified as:
```
if [ "$NGINX_LOG_FORMAT_TIDY_IN_CONTAINER" != "1" ]; then
```

bin/nginx-log-format-tidy.sh
```diff
@@ -0,0 +1,126 @@
...
+    repo_root=$(git rev-parse --show-toplevel)
+
+    if ! command -v docker >/dev/null 2>&1; then
+        >&2 echo 'docker is required to run the nginx log format clang-tidy check.'
+        exit 1
+    fi
+
+    exec docker run --rm -t \
+        -e NGINX_LOG_FORMAT_TIDY_IN_CONTAINER=1 \
+        -e BUILD_TYPE="${BUILD_TYPE:-}" \
```
xlamorlette-datadog: _nit_: This can be simplified as follows (by the way, sorted alphabetically):
```
        -e BUILD_DIR \
        -e BUILD_TYPE \
        -e MAKE_JOB_COUNT \
        -e NGINX_CONF_ARGS \
        -e NGINX_LOG_FORMAT_TIDY_BUILD_DIR \
        -e NGINX_VERSION \
        -e RUM \
        -e WAF \
```

bin/nginx-log-format-tidy.sh
```diff
@@ -0,0 +1,126 @@
...
+    perl \
+    python3 \
+    zlib-dev
+
+jobs=${MAKE_JOB_COUNT:-}
+if [ -z "$jobs" ]; then
+    jobs=$(getconf _NPROCESSORS_ONLN 2>/dev/null || echo 2)
+fi
+
+build_dir=${BUILD_DIR:-.clang-tidy-build/alpine-3.23.4/project}
```
xlamorlette-datadog: _nit_: factorize alpine version number (lines 26, 54 and 55)

bin/nginx-log-format-tidy.sh
```diff
@@ -0,0 +1,126 @@
...
+plugin_build_dir=${NGINX_LOG_FORMAT_TIDY_BUILD_DIR:-.clang-tidy-build/alpine-3.23.4/plugin}
+nginx_version=${NGINX_VERSION:-1.31.1}
```
xlamorlette-datadog: _opt_: rather output an error if `NGINX_VERSION` is not set. I suggest this because when updating the Nginx versions, one may wrongly think that it is needed to update this occurrence.

src/common/headers.cpp
```diff
@@ -1,5 +1,7 @@
 #include "headers.h"
 
+#include <memory>
```
xlamorlette-datadog, line 3: _opt, personal preference_: I prefer to avoid including things already indirectly included, because it is difficult to determine where to stop when including all what is used, and because it adds useless lines of code. I know this is debatable ;-)

.gitignore
```diff
@@ -1,6 +1,7 @@
 /.build
 /.musl-build
 /.openresty-build
+/.clang-tidy-build
```
xlamorlette-datadog: _nit, opt, super pedantic_: keep alphabetically sorted

tools/clang-tidy/README.md
xlamorlette-datadog, line 1: Is the code in this directory original code or what is copied from somewhere?
If copied (or adapted) from a source, please reference it.
If this is original code, I'm very impressed!!! And I would need some more time to review it.

xlamorlette-datadog (approved):
This tool is really great, and the number of fixes is impressive!

src/common/headers.cpp
cataphract: I added this because clang tidy complained otherwise. Probably the toolchain with clang in the alpine image I used doesn't pull the include transitively.

tools/clang-tidy/README.md
cataphract: He, it was mostly written by GPT-5.5. I didn't learn the clang/llvm api just for this :p

Most adjustments I made were functional -- selecting how strictly to match each type. For instance, if we pass `long` to a specifier that requires `ngx_int_t` (which ultimately resolves to `long` on arm64/amd64 linux), should it be accepted? In general, we want to avoid this kind of stuff, because it makes the code non-portable.

xlamorlette-datadog: That's really impressive!!!

PR #401, opened 2026-06-11
fix resource renaming not enabled if appsec enabled via env

cataphract:
When resource renaming is not explicitly enabled or disabled, it should get enabled if appsec is explicitly enabled (not via remote config). However, we were only checking for the enablement via nginx config file, not the env var. Fix that

### DIFF
--- a/test/cases/endpoint_renaming/test_endpoint_renaming.py
+++ b/test/cases/endpoint_renaming/test_endpoint_renaming.py
@@ -24,36 +24,33 @@ class TestEndpointRenaming(case.TestCase):
             self.assertEqual(0, status, log_lines)
             TestEndpointRenaming.last_config = new_config
 
-    def send_request_and_get_span(self, url, config_name):
-        self.replace_config(config_name)
+    def send_request_and_get_span(
+            self,
+            url: str,
+            config_name: str,
+            extra_env: dict[str, str] | None = None) -> tuple[int, dict]:
         self.orch.sync_service('agent')
 
-        status, _, _ = self.orch.send_nginx_http_request(url)
-
-        self.orch.reload_nginx()
-        log_lines = self.orch.sync_service('agent')
-
-        spans = formats.parse_spans(log_lines)
-        nginx_spans = [s for s in spans if s.get('service') == 'test-service']
-
-        self.assertEqual(len(nginx_spans), 1, "Expected exactly one span")
-
-        return status, nginx_spans[0]
-
-    def send_request_and_get_span_with_env(self, url, config_name, extra_env):
-        conf_path = Path(__file__).parent / "conf" / config_name
-        conf_text = conf_path.read_text()
-
-        self.orch.sync_service('agent')
-        try:
-            with self.orch.custom_nginx(conf_text,
-                                        extra_env=extra_env,
-                                        healthcheck_port=80):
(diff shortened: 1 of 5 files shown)


test/cases/endpoint_renaming/test_endpoint_renaming.py
```diff
@@ -156,3 +186,21 @@ def test_appsec_enables_fallback_mode_by_default(self):
             'http.endpoint', meta,
             "http.endpoint should NOT be set when http.route exists in fallback mode (appsec enabled)"
         )
+
+    def test_appsec_env_enables_fallback_mode_by_default(self):
+        """Verify that DD_APPSEC_ENABLED=true enables endpoint renaming by default."""
+        if self.waf_disabled:
```
xlamorlette-datadog: _opt_: Use a decorator to factorize this:

In `case.py`:
```
import functools

def skip_if_waf_disabled(func):
    @functools.wraps(func)
    def wrapper(self, *args, **kwargs):
        if self.waf_disabled:
            self.skipTest("WAF is disabled - appsec test requires WAF support")
        return func(self, *args, **kwargs)
    return wrapper
```

Then:
```
@case.skip_if_waf_disabled
def test_appsec_enables_fallback_mode_by_default(self):

@case.skip_if_waf_disabled
def test_appsec_env_enables_fallback_mode_by_default(self):
```

test/cases/endpoint_renaming/test_endpoint_renaming.py
```diff
@@ -40,6 +40,36 @@ def send_request_and_get_span(self, url, config_name):
 
         return status, nginx_spans[0]
 
+    def send_request_and_get_span_with_env(self, url, config_name, extra_env):
```
xlamorlette-datadog: _nit_: please add type hints for function parameters (and return values)

test/cases/endpoint_renaming/test_endpoint_renaming.py
```diff
@@ -40,6 +40,36 @@ def send_request_and_get_span(self, url, config_name):
...
+            and s.get('meta', {}).get('http.url', '').endswith(url)
+        ]
+
+        self.assertEqual(len(nginx_spans), 1, {
+            "url": url,
+            "spans": nginx_spans,
+            "log_lines": log_lines,
+        })
+
+        return 200, nginx_spans[0]
```
xlamorlette-datadog: _nit_: rather return `status` (even if always `200` because of the assert line 54)?

test/cases/endpoint_renaming/test_endpoint_renaming.py
```diff
@@ -40,6 +40,36 @@ def send_request_and_get_span(self, url, config_name):
...
+                                        healthcheck_port=80):
+                self.orch.sync_service('agent')
+                status, _, _ = self.orch.send_nginx_http_request(url)
+                self.assertEqual(200, status)
+        finally:
+            TestEndpointRenaming.last_config = ''
+
+        log_lines = self.orch.sync_service('agent')
+        spans = formats.parse_spans(log_lines)
+        nginx_spans = [
```
xlamorlette-datadog, line 57: _nit_: should we have a similar spans filtering in `send_request_and_get_span()` above?

xlamorlette-datadog (approved):

PR #394, opened 2026-06-02
Fix two bugs in logging statements

cataphract:
See [APMSP-3398](https://datadoghq.atlassian.net/browse/APMSP-3398) and [APMSP-3399](https://datadoghq.atlassian.net/browse/APMSP-3399)

[APMSP-3398]: https://datadoghq.atlassian.net/browse/APMSP-3398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APMSP-3399]: https://datadoghq.atlassian.net/browse/APMSP-3399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #410, opened 2026-07-16
Fix potential request user-after-free in AppSec tasks

cataphract:
## What does this change?

  Keep nginx requests alive while asynchronous WAF tasks are running.

  ## Why is this needed?

  WAF task submission incremented `r->main->count`, but not `r->main->blocked`.

During request termination, nginx may discard the request count and destroy the request pool when `blocked == 0`. Because  the WAF task and its completion state contain pool-owned data, the task could then complete against freed memory, causing request corruption or a worker crash.

  ## How is it fixed?

  - Increment both `r->main->count` and `r->main->blocked` before posting a WAF task.
  - Roll back both references if task submission fails.
  - Decrement `blocked` when completion returns to nginx’s main thread.
  - If the request was terminated while the task was running, release the task’s count reference and invoke the connection write handler so nginx can finish deferred termination.
  - Do not resume normal WAF processing for an already terminated request.

## Testing

It is not easy to force a request termination in test code, especially given that we cut most paths that would lead to it by replacing either the request or the connection read/write handlers. However, it remains a possibility -- anything calling `ngx_http_finalize_request` e.g. from a timer would trigger the bug. For testing, we add special testing flags that inject delays in the task and start a timer that calls `ngx_http_finalize_request`.

### DIFF
--- a/src/security/context.cpp
+++ b/src/security/context.cpp
@@ -1,9 +1,11 @@
 #include "context.h"
 
 #include <atomic>
+#include <chrono>
 #include <climits>
 #include <optional>
 #include <stdexcept>
+#include <thread>
 #include <type_traits>
 #include <utility>
 
@@ -153,6 +155,7 @@ class PolTaskCtx {
     as_self().replace_handlers();
 
     req_.main->count++;
+    req_.main->blocked++;
 
     bool simulate_task_post_failure = false;
     auto *main_conf = static_cast<datadog_main_conf_t *>(
@@ -172,6 +175,7 @@ class PolTaskCtx {
       Stats::task_submission_failed();
 
       req_.main->count--;
+      req_.main->blocked--;
       as_self().restore_handlers();
 
       as_self().~Self();
@@ -179,6 +183,8 @@ class PolTaskCtx {
     }
 
     Stats::task_submitted();
+
+    maybe_schedule_test_termination(main_conf);
 
     ngx_log_debug2(NGX_LOG_DEBUG_HTTP, req_log(), 0,
                    "task %p submitted. Request refcount: %d", &get_task(),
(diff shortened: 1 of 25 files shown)


src/security/context.cpp
```diff
@@ -203,6 +209,33 @@ class PolTaskCtx {
...
+      terminate_test_event_.data = this;
+      terminate_test_event_.log = req_log();
+      ngx_add_timer(&terminate_test_event_,
+                    main_conf->appsec_test_task_termination_delay_ms);
+    }
+  }
+
+  // runs on the main thread; test-only forced-termination hook
+  static void test_terminate_request(ngx_event_t *evt) noexcept {
+    auto *self = static_cast<PolTaskCtx *>(evt->data);
```
xlamorlette-datadog, line 230: _opt, nit_: I prefer avoiding usage of `auto` for better readability

src/security/context.cpp
```diff
@@ -203,6 +209,33 @@ class PolTaskCtx {
...
+    ngx_connection_t *connection = self->req_.connection;
+
+    ngx_log_debug0(NGX_LOG_DEBUG_HTTP, self->req_log(), 0,
```
xlamorlette-datadog: _nit_: Rather use `ngx_log_debug()` variadic macro (as done in https://github.com/DataDog/nginx-datadog/pull/394).
Same below.

test/cases/sec_blocking/test_task_teardown_race.py
```diff
@@ -0,0 +1,161 @@
...
+    TERMINATION_MASKS = {
+        'test_initial_waf_task_survives_request_termination': INITIAL_WAF_TASK,
+        'test_request_body_waf_task_survives_request_termination':
+        REQUEST_BODY_WAF_TASK,
+        'test_final_waf_task_survives_request_termination': FINAL_WAF_TASK,
+    }
+
+    def setUp(self):
+        super().setUp()
+        waf_path = Path(__file__).parent / './conf/waf.json'
```
xlamorlette-datadog: _nit, opt, super minor_: `./` is superfluous.

xlamorlette-datadog (approved):
Great!

src/security/context.cpp
cataphract: In general, I agree. But when assigning the result of casts, I see no point in repeating the type on the left side of the assignment.

src/security/context.cpp
cataphract: I'm fixing all occurrences in the codebase and adding a clang-tidy check for it. See https://github.com/DataDog/nginx-datadog/pull/410/commits/353a839f916d521f191d7415be11d53e3a9d1454

cataphract:
Added commits to fix the unsigned images problems.

pawelchcki:
I merged by habit, but it DID look good to me!

PR #412, opened 2026-07-16
Avoid loading on ubuntu debian

cataphract:
Debian/Ubuntu builds have an incompatible ABI, in fact, they seem to break the ABI even after security updates. They also don't respect the nginx convention for detecting ABI compatibility and have their own.

Therefore, reject loading on Debian/Ubuntu builds unless the module was built against Debian/Ubuntu sources (e.g. the nginx-dev package).

Additionally, have the Makefile propagate NGINX_SRC_DIR to cmake and cmake internally propagate it to the docker container.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -225,6 +225,12 @@ target_sources(ngx_http_datadog_objs
     src/tracing_library.cpp
 )
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  target_compile_definitions(ngx_http_datadog_objs PRIVATE
+    DD_NGINX_DEBIAN_UBUNTU_BUILD=${NGINX_DEBIAN_UBUNTU_BUILD})
+  target_sources(ngx_http_datadog_objs PRIVATE src/nginx_package_abi.cpp)
+endif()
+
 if(NGINX_DATADOG_ASM_ENABLED)
   target_sources(ngx_http_datadog_objs
     PRIVATE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,7 @@ include(./cmake/deps/rapidjson.cmake)
 include(./cmake/deps/cppcodec.cmake)
 
 add_subdirectory(./dd-trace-cpp EXCLUDE_FROM_ALL)
+set_property(TARGET dd-trace-cpp-static PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 if(ENABLE_ASAN OR ENABLE_MSAN)
   foreach(target_name IN ITEMS
--- a/src/nginx_package_abi.cpp
+++ b/src/nginx_package_abi.cpp
@@ -1,0 +1,155 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
+#include "nginx_package_abi.h"
+
+#include <dlfcn.h>
+#include <sys/mman.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include <array>
+#include <cerrno>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+
+#ifndef DD_NGINX_DEBIAN_UBUNTU_BUILD
+#define DD_NGINX_DEBIAN_UBUNTU_BUILD 0
+#endif
+
+namespace datadog::nginx::package_abi {
+namespace {
+
+using NginxMain = int (*)(int, char *const *);
+
+constexpr std::string_view kUbuntuVersionMarker = "(Ubuntu)";
+constexpr std::string_view kUbuntuConfigureMarker = "--build=Ubuntu";
+constexpr std::string_view kDebianBuildPath =
+    "-ffile-prefix-map=/build/reproducible-path/nginx-";
+constexpr std::array<std::string_view, 3> kDebianPackagingLayout = {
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -225,6 +225,12 @@ target_sources(ngx_http_datadog_objs
     src/tracing_library.cpp
 )
 
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  target_compile_definitions(ngx_http_datadog_objs PRIVATE
+    DD_NGINX_DEBIAN_UBUNTU_BUILD=${NGINX_DEBIAN_UBUNTU_BUILD})
+  target_sources(ngx_http_datadog_objs PRIVATE src/nginx_package_abi.cpp)
+endif()
```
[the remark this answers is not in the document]
cataphract: First, a premise is wrong: in musl, there is no separate libdl and libdl.a is just an empty stub.

The proposed change seems unnecessary. `dlsym` will obviously be present at runtime because it's part of the mechanism to load the module. At static link time it's another matter, I guess it depends whether undefined symbols are errors or result in imports. I'll check in a glibc build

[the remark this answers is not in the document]
cataphract: libdl was consolidated into libc in glibc 2.34 (2021). So build errors would occur, if at all, only against older glibc. OTOH, we want to avoid the unnecessary NEEDED: libdl at runtime, so leaving it out is prob the best option

src/tracing/directives.h
```diff
@@ -45,8 +45,13 @@ char *set_datadog_sample_rate(ngx_conf_t *cf, ngx_command_t *command,
 char *set_datadog_propagation_styles(ngx_conf_t *cf, ngx_command_t *command,
                                      void *conf) noexcept;
 
+#if defined(__clang__)
```
xlamorlette-datadog: _nit_: These lines are duplicated in `src/security/directives.h`, `src/tracing/directives.h` and `src/ngx_http_datadog_module.cpp`. Here is a factorization suggestion, to be checked:

`src/common/directives.h`:
```
#if defined(__clang__)
#define PRAGMA_PUSH_IGNORE_INVALID_OFFSETOF \
  _Pragma("clang diagnostic push") \
  _Pragma("clang diagnostic ignored \"-Winvalid-offsetof\"")
#define PRAGMA_POP_IGNORE_INVALID_OFFSETOF _Pragma("clang diagnostic pop")
#elif defined(__GNUC__)
#define PRAGMA_PUSH_IGNORE_INVALID_OFFSETOF \
  _Pragma("GCC diagnostic push") \
  _Pragma("GCC diagnostic ignored \"-Winvalid-offsetof\"")
#define PRAGMA_POP_IGNORE_INVALID_OFFSETOF _Pragma("GCC diagnostic pop")
#endif
```

`src/security/directives.h`, `src/tracing/directives.h` and `src/ngx_http_datadog_module.cpp`:
```
PRAGMA_PUSH_IGNORE_INVALID_OFFSETOF
[…]
PRAGMA_POP_IGNORE_INVALID_OFFSETOF
```

src/nginx_package_abi.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <string_view>
+
+namespace datadog::nginx::package_abi {
+
+bool is_debian_or_ubuntu_build(std::string_view version_info) noexcept;
+bool running_nginx_is_debian_or_ubuntu() noexcept;
```
xlamorlette-datadog: _nit, opt_: renaming suggestion: I prefer `bool` functions starting with a status verb (`is`, `has`, etc.):
- `bool running_nginx_is_debian_or_ubuntu()` → `bool is_running_nginx_is_debian_or_ubuntu()`
- `bool module_was_built_for_debian_or_ubuntu()` → `bool is_module_built_for_debian_or_ubuntu()`

src/nginx_package_abi.cpp
```diff
@@ -0,0 +1,155 @@
...
+#ifndef DD_NGINX_DEBIAN_UBUNTU_BUILD
+#define DD_NGINX_DEBIAN_UBUNTU_BUILD 0
+#endif
+
+namespace datadog::nginx::package_abi {
+namespace {
+
+using NginxMain = int (*)(int, char *const *);
+
+constexpr std::string_view kUbuntuVersionMarker = "(Ubuntu)";
```
xlamorlette-datadog, line 28: _nit, code style_: no Hungarian prefix and snake case ([example](https://github.com/DataDog/nginx-datadog/blob/4c4df0f/src/tracing/directives.h#L28)):
- `kUbuntuVersionMarker` → `ubuntu_version_marker`
- `kUbuntuConfigureMarker` → `ubuntu_configure_marker`
- `kDebianBuildPath` → `debian_build_path`
- `kDebianPackagingLayout` → `debian_packaging_layout`

xlamorlette-datadog (approved):

src/nginx_package_abi.cpp
cataphract: This is the convention being used in src/security; see https://github.com/DataDog/nginx-datadog/blob/master/src/security/.clang-tidy#L14

xlamorlette-datadog: Thanks! That's really great to have such a `.clang-tidy` configuration file! But it's a shame we have divergent conventions across the project… I'm going to add this in my (of course very long) todo list!…

PR #430, opened 2026-09-10
test: add full nginx injection tests and runnable examples

pawelchcki:
Add installer acceptance tests that verify real Nginx → Flask traces on a simulated Linux host and in ordinary Debian, Alpine, and single-node Docker Swarm workloads. The suite checks HTTP responses, decoded spans, parent relationships, environment settings, agent routing, sampling, disablement, reloads, injection failures, stable configuration, and global RUM opt-out.

The installer configures host or Docker injection. Workload containers and the Swarm service specify no runtime and contain no manual tracing setup. Published packages and checkout-built OCI packages share the same tests, including installed-module checksum verification for checkout artifacts.

Stable-config coverage follows the system-tests YAML schemas for local and managed files. It checks defaults, managed > environment > local precedence, targeting rules, tags, tracing disablement, and global RUM opt-out. The missing C++ tracing support and stable-config RUM opt-out are strict expected failures. An unexpected pass fails CI so the markers must be removed when support lands.

The single-node Swarm smoke test starts an ordinary Nginx service with one replica. It requires a loaded module, an unchanged HTML response with `DD_RUM_ENABLED=false`, and exactly one decoded `nginx.request` span. Artifacts record the node, service, task, container, process map, logs, and span before cleanup. Multi-node scheduling, overlay networking, routing mesh, rolling updates, rescheduling, and node failure remain outside this test.

This reuses the stable-config contract from system-tests. The Nginx-specific execution is separate coverage: installer injection, ordinary host and Docker services, module loading, HTTP responses, decoded spans, trace parenting, and RUM-disabled HTML with tracing still active.

Add architecture-matched GitLab jobs, diagnostic artifacts and JUnit, plus `make test-injection` and host/Docker example commands.

Validation:

- Repository lint and all 10 harness unit tests pass; 83 acceptance cases collect.
- Native ARM64 Orb runs exercised stable config with the published package across host, Debian Docker, and Alpine Docker. The tracing slice produced 18 strict xfails. The files were visible in each workload, and the injected Python backend consumed them, confirming the failures are specific to Nginx tracing support.
- The RUM opt-out slice produced 3 passes for `DD_RUM_ENABLED=false` in the environment and 6 strict xfails for the equivalent local and managed stable-config setting. Every case required an injected module, a real `text/html` response, and a valid Nginx span, so tracing remained active.
- The Swarm smoke test passed on native ARM64 with both the published package and a checkout-built OCI package. Each run saved the Swarm service/task IDs, unchanged HTML, loaded-module evidence, and the decoded `nginx.request` span.
- Both examples served requests, saved linked Nginx → Flask traces, and cleaned up successfully.
- Full native ARM64 Orb runs passed the implemented tracing coverage with both published and checkout-built packages. Each run saved verified Nginx → Flask trace pairs.
- The existing unittest runner discovers all 277 cases without importing pytest; a regression test covers that separation.
- The four injection jobs passed in [pipeline 136708859](https://gitlab.ddbuild.io/DataDog/apm-reliability/nginx-datadog/-/pipelines/136708859): published and checkout-built packages on AMD64 and ARM64. That baseline predates the added Swarm smoke case; the current PR pipeline will run it in each injection job.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,7 @@
 __pycache__/
 .DS_Store
 *.so*
+/test/.venv/
+/test/injection/artifacts/
+/.pytest_cache/
+/test/.pytest_cache/
--- a/test/injection/README.md
+++ b/test/injection/README.md
@@ -3,8 +3,9 @@
 These tests install the real Datadog injector and language packages into a disposable
 Linux host. They check HTTP responses, decoded spans, and Nginx → Flask parent links.
 The same cases run against host processes, a Debian Nginx container, and an Alpine
-Nginx container. Kubernetes is outside this suite. Stable-config tracing cases are
-strict expected failures until the C++ tracer supports the feature.
+Nginx container. A focused smoke test also runs an ordinary Nginx service in a
+single-node Docker Swarm. Kubernetes is outside this suite. Stable-config tracing
+cases are strict expected failures until the C++ tracer supports the feature.
 
 ## Run
 
@@ -36,6 +37,13 @@ To investigate one case:
 DOCKER_CONTEXT=orbstack make test-injection \
   INJECTION_TEST_ARGS='--injection-mode docker-alpine -k test_proxy_trace_link -x' \
   INJECTION_ARTIFACTS=test/injection/artifacts/debug
+```
+
+To run the Swarm smoke test:
+
+```sh
+DOCKER_CONTEXT=orbstack make test-injection \
+  INJECTION_TEST_ARGS='--injection-mode docker-debian -k test_swarm_tracing_without_rum -x'
 ```
 
 Tests run serially. Do not use pytest-xdist. A full run covers all three modes;
@@ -136,6 +144,14 @@ RUM opt-out tests serve HTML while requiring a valid Nginx trace. They verify th
 `DD_RUM_ENABLED=false` in the environment leaves tracing active. The equivalent
 local and managed stable-config cases are strict expected failures until the Nginx
 RUM integration consumes that setting.
+
+The Swarm smoke test creates one service replica in the private Docker daemon. It
+uses the ordinary Nginx image and does not select a runtime or add Datadog Nginx
+directives. It requires a loaded module, one decoded Nginx span, and unchanged HTML
+when RUM injection is disabled. It records the Swarm node, service, task, container,
+logs, process map, and trace before removing the service. This test does not cover
+multi-node scheduling, overlay networking, routing mesh, rolling updates, or node
+failure.
 
 Each known gap uses `xfail(strict=True)`. When support lands, an unexpected pass
 fails CI and requires removing the marker.
--- a/test/injection/README.md
+++ b/test/injection/README.md
@@ -40,10 +40,10 @@ DOCKER_CONTEXT=orbstack make test-injection \
 
 Tests run serially. Do not use pytest-xdist. A full run covers all three modes;
 filtering is for local diagnosis. Skipped cases and unexpected passes fail the run.
-Stable-config cases are marked as strict xfails. Positive conditions have a
-30-second deadline. Negative checks follow graceful workload shutdown and observe
-a three-second quiet period while checking collector health. Readiness uses
-`/ready`; assertions match unique request URIs.
+Known stable-config gaps are marked as strict xfails. Positive conditions have a
+30-second deadline. Negative checks follow graceful workload shutdown and observe a
+three-second quiet period while checking collector health. Readiness uses `/ready`;
+assertions match unique request URIs.
 
 ## What gets installed
 
@@ -132,9 +132,13 @@ targeting rules, service tags, and tracing disablement. These cases follow the
 system-tests YAML shapes for `apm_configuration_default` and
 `apm_configuration_rules`.
 
-Each case uses `xfail(strict=True)`. The current missing C++ tracer feature is an
-expected failure. When support lands, an unexpected pass fails CI and requires
-removing the marker.
+RUM opt-out tests serve HTML while requiring a valid Nginx trace. They verify that
+`DD_RUM_ENABLED=false` in the environment leaves tracing active. The equivalent
+local and managed stable-config cases are strict expected failures until the Nginx
+RUM integration consumes that setting.
+
+Each known gap uses `xfail(strict=True)`. When support lands, an unexpected pass
+fails CI and requires removing the marker.
 
 ## Evidence and CI
 
--- a/test/injection/README.md
+++ b/test/injection/README.md
@@ -3,7 +3,8 @@
 These tests install the real Datadog injector and language packages into a disposable
 Linux host. They check HTTP responses, decoded spans, and Nginx → Flask parent links.
 The same cases run against host processes, a Debian Nginx container, and an Alpine
-Nginx container. Kubernetes and stable config are outside this suite.
+Nginx container. Kubernetes is outside this suite. Stable-config tracing cases are
+strict expected failures until the C++ tracer supports the feature.
 
 ## Run
 
@@ -38,10 +39,11 @@ DOCKER_CONTEXT=orbstack make test-injection \
 ```
 
 Tests run serially. Do not use pytest-xdist. A full run covers all three modes;
-filtering is for local diagnosis. Skipped cases fail the run. Positive conditions
-have a 30-second deadline. Negative checks follow graceful workload shutdown and
-observe a three-second quiet period while checking collector health. Readiness
-uses `/ready`; assertions match unique request URIs.
+filtering is for local diagnosis. Skipped cases and unexpected passes fail the run.
+Stable-config cases are marked as strict xfails. Positive conditions have a
+30-second deadline. Negative checks follow graceful workload shutdown and observe
+a three-second quiet period while checking collector health. Readiness uses
+`/ready`; assertions match unique request URIs.
 
 ## What gets installed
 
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -7,3 +7,7 @@
 __pycache__/
 .DS_Store
 *.so*
+/test/.venv/
+/test/injection/artifacts/
+/.pytest_cache/
```
xlamorlette-datadog, line 12: _nit_: sort

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
+# Nginx injection acceptance tests
```
xlamorlette-datadog, line 1: _nit_: capitalize titles

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+These tests install the real Datadog injector and language packages into a disposable
```
xlamorlette-datadog, line 3: _nit_: opening paragraph leads with mechanics, not purpose, and uses informal terminology.

- "Datadog injector" isn't the official name — the package is datadog-apm-inject, and Datadog's docs call this "APM auto-instrumentation" / "Single Step Instrumentation." Worth using that term (or naming the package) so readers unfamiliar with this repo's shorthand aren't confused.
- The paragraph opens with how ("install the injector and language packages into a disposable host") before saying what this suite verifies. Suggest leading with the objective — e.g., "These tests verify that Datadog's auto-instrumentation correctly activates nginx-datadog tracing end-to-end" — then the mechanism (host + Debian/Alpine containers + Swarm smoke test), then the caveats (Kubernetes out of scope, stable-config xfails).

test/injection/harness.py
```diff
@@ -0,0 +1,786 @@
...
+    def environment(self, overrides):
+        return {
+            "DD_TRACE_AGENT_URL": "http://127.0.0.1:8126",
+            "DD_APM_INSTRUMENTATION_DEBUG": "true",
+            "DD_TRACE_STARTUP_LOGS": "false",
+            **overrides
+        }
+
+    def start_host(self, service, command, overrides):
+        script = f"echo $$ > /tmp/{service}-launcher.pid; exec {shlex.join(command)} " \
```
xlamorlette-datadog, line 499: This seems useless: this writes `/tmp/{service}-launcher.pid` for Nginx, but `shutdown()` (line 607) reads Nginx's own `pidfile` instead.

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+Install Docker and `uv`, then run from the repository root:
+
+```sh
+DOCKER_CONTEXT=orbstack make test-injection INJECTION_SOURCE=published
+DOCKER_CONTEXT=orbstack make test-injection INJECTION_SOURCE=checkout
+make test-injection INJECTION_PACKAGE=/path/to/package.tar
+```
+
+The Docker daemon must support privileged Linux containers. Orb uses its native ARM64
```
xlamorlette-datadog, line 20: It seems that `Images.__init__()` detects arch dynamically and CI runs both amd64 and arm64.

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
...
+make injection-example-traces
+make injection-example-down
+
+DOCKER_CONTEXT=orbstack make injection-example-up INJECTION_MODE=docker
+make injection-example-request
+make injection-example-traces
+make injection-example-down
+```
+
+`docker-alpine` selects Alpine. `INJECTION_SOURCE=checkout` and `INJECTION_PACKAGE`
```
xlamorlette-datadog, line 96: _nit_:  This paragraph looks like pretty unrelated facts: image selection, source overrides, state-file persistence, artifact-path consistency, how requests work, what shutdown does. I think it lacks something to orient the reader.

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
...
+| Disable Nginx reporting | `DD_TRACE_ENABLED=false` |
+| Opt Nginx out of injection | `DD_INSTRUMENT_SERVICE_WITH_APM=false` |
+
+`DD_TRACE_AGENT_URL` takes precedence over host/port. Collector A listens on 8126
+and the Unix socket; collector B listens on 9126. These addresses are inside the
+disposable host. Flask remains traced when Nginx reporting or injection is disabled.
+The sampling test checks the backend's received priority of `-1`; absence of raw
+traces alone does not establish a sampling decision.
+
+Nginx captures supported Datadog environment variables when its master starts and
```
xlamorlette-datadog, line 123: This says reload retains the original env values while restart picks up new ones. But `test_reload_and_restart()` never changes `DD_ENV`/`DD_VERSION` between steps (it only confirms the same value survives both reload and restart).

test/injection/test_injection.py
```diff
@@ -0,0 +1,481 @@
+import json
+import re
+
+import pytest
+
+from .harness import NGINX_VERSION, STABLE_CONFIG_PATHS, Workload, poll, trace_id, unique_uri, UNSUPPORTED_IMAGE
+
+stable_config_xfail = pytest.mark.xfail(
+    reason="C++ tracer stable configuration support is not implemented",
+    strict=True)
```
xlamorlette-datadog, line 10: I seems that `strict=True` is already global (see ` pyproject.toml` line 15).

test/injection/README.md
```diff
@@ -0,0 +1,177 @@
...
+logs, process map, and trace before removing the service. This test does not cover
+multi-node scheduling, overlay networking, routing mesh, rolling updates, or node
+failure.
+
+Each known gap uses `xfail(strict=True)`. When support lands, an unexpected pass
+fails CI and requires removing the marker.
+
+## Evidence and CI
+
+Artifacts default to `test/injection/artifacts/latest/`. They include JUnit,
```
xlamorlette-datadog, line 161: Could you split this into a short bullet list?

test/injection/package.py
```diff
@@ -0,0 +1,152 @@
...
+                        checksums.append(
+                            hashlib.sha256(
+                                contents.extractfile(
+                                    member).read()).hexdigest())
+        assert len(checksums
+                   ) == 1, f"Missing or duplicate Nginx {NGINX_VERSION} module"
+        return checksums[0]
+
+
+def build(docker, arch):
```
xlamorlette-datadog, line 69: `build()` mixes four distinct concerns: pulling/starting the toolchain container, packaging the repo into a tarball (with its own nested include filter) and copying it in, running the musl build, then extracting the module and computing the version. I think this could be split, for example, into helpers, so that `build()` would then read as a short pipeline instead of one long procedural block.

test/injection/run.py
```diff
@@ -0,0 +1,160 @@
...
+import os
+from pathlib import Path
+import subprocess
+import sys
+
+from .harness import Docker, Images, MODES, ROOT, Sandbox, Workload, unique_uri
+from .package import build, module_checksum
+
+
+def main():
```
xlamorlette-datadog, line 15: CLI parsing, arch detection, and three separate package-resolution strategies (explicit `--package`, glob a CI artifact, or build from checkout) are all inline before dispatch even happens. I think one should extract things such as `resolve_package(args, docker, arch)` so that `main()` would shrink to parse → resolve → dispatch, and the resolution logic could be tested on its own.

test/injection/harness.py
```diff
@@ -0,0 +1,786 @@
...
+        self.sandbox.host("kill", "-HUP", str(master))
+        self.generation += 1
+        poll(
+            lambda: {
+                item["pid"]
+                for item in self.evidence()
+                if "worker process" in item["command"]
+            }.isdisjoint(workers), "new Nginx workers")
+
+    def stop(self):
```
xlamorlette-datadog, line 600: `stop()` branches into two long, unrelated shutdown procedures depending on `sandbox.mode`. I think you could extract helpers such as `_stop_host()` /  `_stop_docker()`, so that conditional blocks don't hide they are two different implementations of the same idea.

test/injection/test_injection.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+@pytest.mark.parametrize("failure", ["missing-package", "unsupported-version"])
+def test_unavailable_module(sandbox_factory, mode, failure):
+    sandbox = sandbox_factory(mode, f"{mode}-{failure}")
+    try:
+        sandbox.start()
+        sandbox.install(nginx=failure != "missing-package")
+        image = None
+        if failure == "unsupported-version":
```
xlamorlette-datadog, line 443: This branch does something quite specific (copying an old nginx binary and its libraries out of a container to fake an unsupported host install) nested inside a test that's otherwise about asserting injection was skipped. One could extract that setup into something as `_install_unsupported_host_nginx(sandbox, image)` so the test body reads as: set up failure scenario → request → assert skipped-injection diagnostics.

xlamorlette-datadog (changes requested):
Thanks for putting this together! The amount of acceptance coverage here (host, Debian, Alpine, Swarm, stable config, RUM opt-out, all validated on real ARM64 runs) is genuinely impressive! Unfortunately, I don't think I can give this the review it deserves in its current shape.

This is 2,526 new lines across 19 files, all additions, with no way to review it incrementally (`harness.py` alone is 786 lines, `test_injection.py` is 481). A change like this needs to be read almost linearly to hold the whole model (Sandbox/Workload lifecycle, host vs. Docker branching, stable-config precedence) in your head at once, and that's a lot to keep loaded while also checking correctness.

I have much difficulties reading the readme. It's the main way I'd normally build a mental model before diving into the code, but several sections pack many unrelated facts into single paragraphs with no topic sentence (for example, lines 96–101 cover image selection, source overrides, state persistence, artifact consistency, request mechanics, and shutdown behavior all in one breath, and lines 159–167 list eleven different artifact types plus cleanup rules the same way). It reads more like a list of notes than a document meant to guide a reader, so I end up having to re-read paragraphs to pull out the one fact I actually need.

The PR description itself has the same issue. It's dense with facts (examples: "18 strict xfails", "83 acceptance cases",  "277 cases",  "pipeline ID 136708859") but they're not laid out so I can check them against anything. I can't easily tell which numbers I'm supposed to verify versus just trust, and some claims describe behavior from a specific ARM64 run I have no way to reproduce or double check (example: "Native ARM64 Orb runs exercised stable config...").

To be honest, so far, I just did same sampling review, not a full review that this work deserve. My concern is purely that, as submitted, it's too large and too densely written for me to review carefully enough to be useful to you.

PR #431, opened 2026-09-10
Move to the public musl-build-env

cataphract:
With https://hub.docker.com/r/datadog/musl-build-env/tags now public, we can move to it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,27 +52,22 @@ IN_DOCKER_OR_CI := $(shell if [ "$(IN_DOCKER)" = "true" ] || \
 
 # ----- Docker Images
 
-BUILD_IMAGE := nginx_musl_toolchain
+BUILD_IMAGE_DIGEST := sha256:6fbb23a5089853eea7cd26dfaeded4e6541027927016fb1ccdaf2d6713da5c26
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,40 +52,31 @@ IN_DOCKER_OR_CI := $(shell if [ "$(IN_DOCKER)" = "true" ] || \
 
 # ----- Docker Images
 
-BUILD_IMAGE := nginx_musl_toolchain
+BUILD_IMAGE_DIGEST := sha256:6fbb23a5089853eea7cd26dfaeded4e6541027927016fb1ccdaf2d6713da5c26
```
xlamorlette-datadog: _nit_: add the version tag as a comment at the end of the line `# 1.0.5`

xlamorlette-datadog: _nit_: Maybe it would be nice to add a comment stating that the build image comes from https://github.com/DataDog/musl-toolchain-glibc-support. What do you think?

xlamorlette-datadog (approved):
The creation of https://github.com/DataDog/musl-toolchain-glibc-support is really great! Thanks a lot!!!

